PR #5, opened 2023-08-29
Feature/reqmgmt add charter

gregshue:
Capture Zephyr Project Charter, dated August 21, 2023, in StrictDoc (.sdoc) format so the project can show how those constraints are being met.

The document is placed in a `stakeholders/` directory to clearly indicate the source and scope of the content.


nicpappler:
I strongly object to add the Zephyr Project's Charter to the requirements. The charter defines the overall structure of the project wrt to the project organisation etc, but no functional nor non-functional requirements that would be of any value to Zephyr as a product.

gregshue:
> but no functional nor non-functional requirements that would be of any value to Zephyr as a product.

Please look more carefully at the following:

- 1.a: secure and safe (functional & quality requirements)
- 1.b: is open source (quality, licensing)
- 6/7.b: definition of process to ensure an auditable codebase (non-functional; possibly functional)
- 6/7.b: [ensure] any associated certification artifacts (functionality of the ecosystem)
- 6/7.c: only Platinum Members will have access to the Security/Safety Artifacts (functionality of the Zephyr Project solution available to members). Note that ZP must provide the tests and test procedures so that any user can produce these artifacts for themselves (otherwise it isn't very useful).

AFAICT , the clauses above result in the Zephyr Project **having** to provide:
1. A qualified/qualifiable build system that supports OOT/user proprietary code (to create a certifiable image).
2. A qualified/qualifiable test framework & tests reusable by the end user (to create certification artifacts).
3. A qualified/qualifiable requirements management system and requirements extensible for downstream users, including chip vendors making drivers reusable in safety-critical systems (to create certification artifacts).
4. A qualified/qualifiable configuration system to select which subset of functionality (or implementations) to be built into an end-user's product (to allow the certification artifacts to be reused for an end-user's product).
6. A qualified/qualifiable solution for describing HW constraints for user-defined SOC, boards, etc. (to allow the certification artifacts to be reused for an end-user's product).

Thoughts?

gregshue:
@nicpappler I also noticed that charter item 14.b is another non-functional requirement on outbound code (Apache 2.0 licensed). There is an analogous requirement for documentation (Creative Commons Attribution 4.0 International License), which may be reused by end users. ;-)

nicpappler:
I don't doubt that we have to create high-level, non-functional requirements. But the project charta is not the right source for it. Think along the lines of assumed requirements and Safety Manuals. Obligations of use. That stuff. The charta is about the Zephyr project organisation, not about the product. It is simply the wrong source.

gregshue:
> The charta is about the Zephyr project organisation, not about the product.

I'm not following you. I just highlighted ~10 technical points from the charter that constrain what must be produced by the Zephyr Project for end use. Additional technical points exist that constraint contributions to the product. Like it or not, the charter _is_ also about the product, and it is a source we must work within.

The charter is a stakeholder-level document (think business sponsor) where stakeholders have put constraints on the product itself (e.g., Apache 2.0). I am not aware of any more authoritative source, and we must honor those explicit constraints. All project decisions should align with them, whether technical or not.

I'm basing this approach on [Requirements Elicitation](http://swebokwiki.org/Chapter_1:_Software_Requirements#Requirements_Elicitation) (which is one of the sources used for [The Linux Foundation's free, self-paced SW Engineering Basics](https://training.linuxfoundation.org/training/software-engineering-basics-for-embedded-systems-lfd116/) course). I'm sure the LF approves of this source material and Requirements Engineering (RE) process.

Has the LF approved of a different RE approach for this project? If so, can you provide a link to it so I can get on board?

nashif:
I think there is general consensus the charter as a whole is not a source for requirements, lets close this to avoid distracting people.

PR #6, opened 2023-09-12
Req: Add Review changes from the safety committee

simhein:
Add changes to requierements that were reviewed and approved by the safety committee.

### DIFF
--- a/docs/zephyr_02_functional_requirements.sdoc
+++ b/docs/zephyr_02_functional_requirements.sdoc
@@ -49,10 +49,10 @@ REFS:
   VALUE: ZEP-ARCH-001
 TITLE: Atomic Operations
 STATEMENT: >>>
-Zephyr shall provide an interface functionality to access memory while ensuring mutual exclusion. Note: Implementation by atomic variables and accessing them by APIs
+Zephyr shall provide an interface functionality to access memory while ensuring mutual exclusion. Note: Implementation by atomic variables and accessing them by APIs.
 <<<
 USER_STORY: >>>
-As a Zephyr OS user I want to read from or write into a memory areas without being disturbed by other threads or ISRs
+As a Zephyr OS user I want to read from or write into a memory areas without being disturbed by other threads or ISRs.
 <<<
 DISCUSSION_DATE: >>>
 2022/3/30 - ok - pq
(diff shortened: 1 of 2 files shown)


stanislaw (commented):
This is a very good cleanup. Only a small stylistic comment from my StrictDoc experience :smile: 

To have it consistent everywhere, you could ensure that all multiline field sentences end with a dot. (STATEMENT and USER_STORY)

(StrictDoc could have an optional validation warning for this in the near future)

simhein:
> This is a very good cleanup. Only a small stylistic comment from my StrictDoc experience 😄
> 
> To have it consistent everywhere, you could ensure that all multiline field sentences end with a dot. (STATEMENT and USER_STORY)
> 
> (StrictDoc could have an optional validation warning for this in the near future)

Thanks for the input, I will go through it again and will provide an extra commit for the fix to be consistent.

docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -308,12 +308,12 @@ COMPONENT: C Library
...
   VALUE: ZEP-CLIB-001
-TITLE: Selected C libraries documentation in the Zephyr Safety Manual
+TITLE: Document set of Zephyr OS required C librariy functions in Safety Manual
 STATEMENT: >>>
-The set of C Libraries supported by Zephyr needs to be defined documented in the Zephyr Safety Manual
+The set of C Library functions required by Zephyr needs to be documented in the Zephyr Safety Manual
 <<<
 USER_STORY: >>>
-As a Zephyr OS user implementing a safety application I want to have documented which C library functions are supported.
+As a Zephyr user, I want to know which C library functions are being called by the Zephyr OS safety scope code.
```
gregshue, line 316: The new statement is a very different scope than the old one.  AFAICT, generating the Safety Artifacts will require ZP to provide a safety-certified implementation of the C library functions used. Documenting the safety-certifiable subset provided is also important for end users.

simhein: In this case the change of the scope in this req. is intended. The thing here is, that there will be no generated artifact because zephyr will always be a part system which need to be certified in a complete state. To minimize the initial certification effort, the c library functions are not in scope right now. Also we don't know what toolchains / libs will be used in the downstream safety-critical systems, we decided that this needs to be taken care of the integrator who wants to build a safety-critical system with zephyr.

PR #10, opened 2023-10-31
 Zephyr high-level and functional requirements: migrate to RELATIONS and auto-generate UIDs 

stanislaw:
This work concludes the assignment of ZEP-based UIDs to all existing Zephyr requirements.

The last auto-generated UID is `ZEP-121` which matches the total number of requirements.

The first two commits update the requirements to be in accordance with the latest StrictDoc conventions: 
- The deprecated `REFS` field has been renamed to the `RELATIONS` field.
- The `RELATIONS` field is now the last field in each requirement.

The third commit auto-generates the remaining UIDs.

The changes can be reviewed commit-by-commit.

PR #11, opened 2023-11-06
Req: Add fix requirements prefix

simhein:
Add the defined requirements prefix (ZEP)
to the document definition


simhein:
Duplicate to PR #10

PR #15, opened 2023-12-05
Thread Requierements

simhein:
- Added new Thread Requirements
- Created new Thread sections
- Moved already written reqs to the threads section
- Removed duplicated and not needed requirements
- Add linking of requirements.

### DIFF
--- a/docs/zephyr_01_high_level_requirements.sdoc
+++ b/docs/zephyr_01_high_level_requirements.sdoc
@@ -272,7 +272,7 @@ pf-ok
 UID: ZEP-4
 STATUS: Draft
 TYPE: High Level
-COMPONENT: Thread Scheduling
+COMPONENT: Scheduling
 TITLE: Scheduling
 STATEMENT: >>>
 Zephyr shall provide an interface to assign a thread to a specific CPU.
@@ -308,7 +308,7 @@ UID: ZEP-5
 STATUS: Draft
 TYPE: High Level
 COMPONENT: Threads
-TITLE: Managing threads
+TITLE: Thread management
 STATEMENT: >>>
 Zephyr shall provide a framework for managing multiple threads of execution.
 <<<
--- a/docs/zephyr_01_high_level_requirements.sdoc
+++ b/docs/zephyr_01_high_level_requirements.sdoc
@@ -287,6 +287,22 @@ REVIEW_COMMENT: >>>
 TBD: The statement contains two separate statements.
 <<<
 
+[SECTION]
+TITLE: Threads
+
+[REQUIREMENT]
+UID: ZEP-123
+STATUS: Draft
+TYPE: High Level
+COMPONENT: Threads
+TITLE: Thread support
+STATEMENT: >>>
+Zephyr shall support threads.
+<<<
+USER_STORY: >>>
+As a Zephyr OS user, I want to be able to have support for the kernel objects named threads for processing work.
+<<<
+
 [REQUIREMENT]
 UID: ZEP-5
 STATUS: Draft
@@ -305,6 +321,21 @@ pf-ok
 REVIEW_COMMENT: >>>
 TBD: Nothing about priorities...
 <<<
+
+[REQUIREMENT]
+UID: ZEP-122
+STATUS: Draft
+TYPE: High Level
+COMPONENT: Threads
+TITLE: Thread priority
+STATEMENT: >>>
+Threads shall have a priority.
+<<<
+USER_STORY: >>>
+As a Zephyr OS user, I want to be able to give my threads different priorities for execution.
+<<<
+
+[/SECTION]
 
 [REQUIREMENT]
 UID: ZEP-6
--- a/docs/zephyr_02_functional_requirements.sdoc
+++ b/docs/zephyr_02_functional_requirements.sdoc
@@ -1303,25 +1303,6 @@ DISCUSSION_DATE: >>>
 <<<
 
 [REQUIREMENT]
-UID: ZEP-98
-STATUS: Draft
-TYPE: Functional
-COMPONENT: Thread Communication
-TITLE: Stack
-STATEMENT: >>>
-The Zephyr kernel shall provide a stack.
-<<<
-USER_STORY: >>>
-As a Zephyr OS user I want to... ???,
-<<<
-DISCUSSION_DATE: >>>
-2022/5/25 - needs to be clarified what stack refers to
-<<<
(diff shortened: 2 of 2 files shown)


simhein:
@stanislaw I can't add you as reviewer, but a short review from your side would also be very welcome :)

docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -1680,23 +1572,165 @@ what does traditional mean here?   Round robin?
...
+UID: ZEP-98
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Threads
+TITLE: Thread stack objects
+STATEMENT: >>>
+Every Thread shall have it's own stack.
+<<<
+USER_STORY: >>>
+As a Zephyr OS user I want to be able to configure the stack size of a thread. And every thread shall have it's own dedicated stack.
```
tobiaskaestner, line 1707: **nitpick:** Whenever I see AND in some RQT it's probably two of them.

docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -1680,23 +1572,165 @@ what does traditional mean here?   Round robin?
...
+  VALUE: ZEP-5
+
+[REQUIREMENT]
+UID: ZEP-113
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Threads
+TITLE: Meta-IRQ Priorities
+STATEMENT: >>>
+The Zephyr kernel shall support running threads in Meta-IRQ priorities.
```
tobiaskaestner, line 1722: **issue:** Without explaining what Meta-IRQ are/mean (as it is not clear from general grounds) this doesn't make sense yet I am afraid. More than that we have two more levels for priorities namely pre-emptible ones (lowest prio, positive prio number) and cooperative ones (of which Meta-IRQs are some specialisation) which we should also consider.

docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -1723,19 +1762,45 @@ Zephyr shall provide an interface to schedule multiple threads.
...
-UID: ZEP-121
+UID: ZEP-125
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Threads
-TITLE: Threads - TBD
+TITLE: Thread Options
 STATEMENT: >>>
-??? Can Zephyr change priviledge level of a thread once created??
+The Zephyr kernel shall support a set of thread options.
```
tobiaskaestner, line 1776: **question:** What is that hinting at? What application level problem does that solve?

docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -1723,19 +1762,45 @@ Zephyr shall provide an interface to schedule multiple threads.
...
+  VALUE: ZEP-123
+- TYPE: Parent
+  VALUE: ZEP-5
+
+[REQUIREMENT]
+UID: ZEP-126
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Threads
+TITLE: Thread Custom Data
```
tobiaskaestner, line 1792: **question:** Is that the same as Thread Local Storage or something different?

tobiaskaestner (commented):
I wasn't sure about the best process so tried the simplest thing that works first. So my comments are inline as PR comments. If we need to change this to be captured in strictdoc itself we probably need to discuss the process, too.

I think starting with the Thread Requirements is a very good choice since everything RTOS basically evolves around the very idea of structuring an application into multiple concurrently running blocks of code. 

That said, I think we can improve the requirements even further to capture crisp and clear the problems Zephyr solves for the application developer and move some jargon to the design specifications, e.g. Meta-IRQs, Kernel objects, etc.

Also, we will probably find more RQTs related to Threads that might not have been included in this first version of the PR.

simhein:
> I wasn't sure about the best process so tried the simplest thing that works first. So my comments are inline as PR comments. If we need to change this to be captured in strictdoc itself we probably need to discuss the process, too.
> 


That is good the Intention was to do it in github. to be open for every one and transparent. What we need to do is write down the process how we want to do this and the acceptance criteria .

docs/zephyr_01_high_level_requirements.sdoc
```diff
@@ -306,6 +322,21 @@ REVIEW_COMMENT: >>>
...
 <<<
 
+[REQUIREMENT]
+UID: ZEP-122
+STATUS: Draft
+TYPE: High Level
+COMPONENT: Threads
+TITLE: Thread priority
+STATEMENT: >>>
+Threads shall have a priority.
```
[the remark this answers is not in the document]
simhein: > **question:** The user should have full control over assigning priorities to threads, right? 
That is right.
And I agree this could be one of a couple RQTs for the scheduling section. But for the Thread itself I would leave it like this or do you prefer to change it?

docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -1680,23 +1572,165 @@ what does traditional mean here?   Round robin?
...
+  VALUE: ZEP-5
+
+[REQUIREMENT]
+UID: ZEP-107
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Threads
+TITLE: Setting thread priority
+STATEMENT: >>>
+Zephyr shall provide an interface to set a thread's priority.
```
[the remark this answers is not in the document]
simhein: I think including it here wouldn't bee good. I think that this information belongs into the design iteself and how the priority is implemented.

docs/zephyr_02_functional_requirements.sdoc
simhein: True but in this case it is in the user story why this RQT exists and not in the statement itself.

docs/zephyr_02_functional_requirements.sdoc
simhein: It solves the option that threads can have special treatment like a user thread or an essential thread.

docs/zephyr_02_functional_requirements.sdoc
simhein: I think it is a little bit different. It needs configuration and is only accessible by the thread itself and it is not available in ISRs

docs/zephyr_02_functional_requirements.sdoc
simhein: True will remove the Meta-IRQ from here because it is described in the design of priorities

nicpappler (approved):

PR #31, opened 2024-03-13
(shelved) gregshue/zep semaphore reqs

gregshue:
Preliminary Low-level Counting Semaphore requirements in EARS syntax. Please note that these requirements are expressed in terms of observable behaviors so are not specific to any implementation. (These requirements could be met by an RTOS implemented in C, C++, or Rust.)  

Requirement statements at this level are sufficient to design verification tests covering full functionality. Once the Counting Semaphore interface is specified and stubs written, the full-functionality verification tests could be implemented, and failure detection verified as the code-under-test evolves to meet requirements.  

Code coverage tools are necessary to identify any unspecified logic and/or data present in the implementation.

### DIFF
--- a/docs/zephyr_02_functional_requirements.sdoc
+++ b/docs/zephyr_02_functional_requirements.sdoc
@@ -1155,6 +1155,23 @@ the Zephyr RTOS shall accomplish all of the following:
 [SECTION]
 TITLE: Semaphore Take
 
+[REQUIREMENT]
+UID: ZEP-129
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Counting Semaphore
+TITLE: Semaphore Take while Available
+STATEMENT: >>>
+While a Counting Semaphore count is > 0,
+
+When a takes that Counting Semaphore,
+
+the Zephyr RTOS shall accomplish all of the following:
+
+- decrement by 1 the count of that Counting Semaphore;
+- indicate SUCCESS to the taking thread.
+<<<
+
 [/SECTION]
 
 [SECTION]
--- a/docs/zephyr_02_functional_requirements.sdoc
+++ b/docs/zephyr_02_functional_requirements.sdoc
@@ -1117,6 +1117,23 @@ TITLE: Initialization
 TITLE: Semaphore Give
 
 [REQUIREMENT]
+UID: ZEP-128
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Counting Semaphore
+TITLE: Semaphore Give without Waiting Threads
+STATEMENT: >>>
+While no thread is waiting for a Counting Semaphore,
+
+When a thread gives that Counting Semaphore,
+
+the Zephyr RTOS shall accomplish all of the following:
+
+- increase the Semaphore Count by 1;
+- indicate SUCCESS to the giving thread.
+<<<
+
+[REQUIREMENT]
 UID: ZEP-127
 STATUS: Draft
 TYPE: Functional
--- a/docs/zephyr_02_functional_requirements.sdoc
+++ b/docs/zephyr_02_functional_requirements.sdoc
@@ -1116,6 +1116,23 @@ TITLE: Initialization
 [SECTION]
 TITLE: Semaphore Give
 
+[REQUIREMENT]
+UID: ZEP-127
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Counting Semaphore
+TITLE: Semaphore Give while Thread Waiting
+STATEMENT: >>>
+While a thread is waiting for a Counting Semaphore,
+
+When a thread gives that Counting Semaphore,
+
+the Zephyr RTOS shall accomplish all of the following:
+
+- unblock the highest priority thread waiting for that Counting Semaphore;
+- indicate SUCCESS to the giving thread.
+<<<
+
 [/SECTION]
 
 [SECTION]
--- a/docs/zephyr_02_functional_requirements.sdoc
+++ b/docs/zephyr_02_functional_requirements.sdoc
@@ -1113,6 +1113,11 @@ TITLE: Initialization
 
 [/SECTION]
 
+[SECTION]
+TITLE: Semaphore Give
+
+[/SECTION]
+
 [/SECTION]
 
 [/SECTION]
--- a/docs/zephyr_02_functional_requirements.sdoc
+++ b/docs/zephyr_02_functional_requirements.sdoc
@@ -1111,6 +1111,19 @@ TITLE: Counting Semaphore
 [SECTION]
 TITLE: Initialization
 
+[REQUIREMENT]
+UID: ZEP-134
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Counting Semaphore
+TITLE: Build-Time Allocation and Initialization
+STATEMENT: >>>
+The Zephyr RTOS shall expose an interface for a Counting Semaphore to accomplish all of the following at build-time:
+
+- statically allocating the Counting Semaphore;
+- statically initialize the Counting Semaphore with the provided count limit and initial count.
+<<<
+
 [/SECTION]
 
 [SECTION]
--- a/docs/zephyr_02_functional_requirements.sdoc
+++ b/docs/zephyr_02_functional_requirements.sdoc
@@ -1219,6 +1219,22 @@ the Zephyr RTOS shall accomplish all of the following:
 - reschedule the waiting thread.
 <<<
 
+[REQUIREMENT]
+UID: ZEP-133
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Counting Semaphore
+TITLE: Semaphore Take Unblocked by Waiting Period Timeout
+STATEMENT: >>>
+When the waiting thread is unblocked by the expiration of the waiting period,
+
+the Zephyr RTOS shall accomplish all of the following:
+
+- resume the waiting thread;
+- indicate TIMEDOUT to the waiting thread;
+- reschedule the waiting thread.
+<<<
+
 [/SECTION]
 
 [SECTION]
--- a/docs/zephyr_02_functional_requirements.sdoc
+++ b/docs/zephyr_02_functional_requirements.sdoc
@@ -1124,6 +1124,19 @@ The Zephyr RTOS shall expose an interface for a Counting Semaphore to accomplish
 - statically initialize the Counting Semaphore with the provided count limit and initial count.
 <<<
 
+[REQUIREMENT]
+UID: ZEP-135
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Counting Semaphore
+TITLE: Dynamic Initialization
+STATEMENT: >>>
+The Zephyr RTOS shall expose an interface to dynamically initialize a Counting Semaphore which requires the following information:
+
+- initial count of the Counting Semaphore;
+- count limit of the Counting Semaphore.
+<<<
+
 [/SECTION]
 
 [SECTION]
--- a/docs/zephyr_02_functional_requirements.sdoc
+++ b/docs/zephyr_02_functional_requirements.sdoc
@@ -1123,6 +1123,11 @@ TITLE: Semaphore Take
 
 [/SECTION]
 
+[SECTION]
+TITLE: Semaphore Count
+
+[/SECTION]
+
 [/SECTION]
 
 [/SECTION]
--- a/docs/zephyr_02_functional_requirements.sdoc
+++ b/docs/zephyr_02_functional_requirements.sdoc
@@ -1108,6 +1108,11 @@ execution explicitly.
 [SECTION]
 TITLE: Counting Semaphore
 
+[SECTION]
+TITLE: Initialization
+
+[/SECTION]
+
 [/SECTION]
 
 [/SECTION]
--- a/docs/zephyr_02_functional_requirements.sdoc
+++ b/docs/zephyr_02_functional_requirements.sdoc
@@ -1266,6 +1266,18 @@ the Zephyr RTOS shall accomplish all of the following:
 [SECTION]
 TITLE: Semaphore Count
 
+[REQUIREMENT]
+UID: ZEP-136
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Counting Semaphore
+TITLE: Current Count
+STATEMENT: >>>
+When the Counting Semaphore current count is requested,
+
+the Zephyr RTOS shall report the current count of the Counting Semaphore.
+<<<
+
 [/SECTION]
 
 [/SECTION]
--- a/docs/zephyr_02_functional_requirements.sdoc
+++ b/docs/zephyr_02_functional_requirements.sdoc
@@ -1105,6 +1105,11 @@ allows for asynchronous communication, where threads do not need to synchronize 
 execution explicitly.
 <<<
 
+[SECTION]
+TITLE: Counting Semaphore
+
+[/SECTION]
+
 [/SECTION]
 
 [SECTION]
--- a/docs/zephyr_02_functional_requirements.sdoc
+++ b/docs/zephyr_02_functional_requirements.sdoc
@@ -1118,6 +1118,11 @@ TITLE: Semaphore Give
 
 [/SECTION]
 
+[SECTION]
+TITLE: Semaphore Take
+
+[/SECTION]
+
 [/SECTION]
 
 [/SECTION]
--- a/docs/zephyr_02_functional_requirements.sdoc
+++ b/docs/zephyr_02_functional_requirements.sdoc
@@ -1172,6 +1172,23 @@ the Zephyr RTOS shall accomplish all of the following:
 - indicate SUCCESS to the taking thread.
 <<<
 
+[REQUIREMENT]
+UID: ZEP-130
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Counting Semaphore
+TITLE: Semaphore Take with TIMEOUT > 0 blocks if not Available
+STATEMENT: >>>
+While a Counting Semaphore count is 0,
+
+When a thread takes that Counting Semaphore with a waiting TIMEOUT > 0,
+
+the Zephyr RTOS shall accomplish all of the following:
+
+- block the taking thread waiting for the availability of that Counting Semaphore;
+- enter the waiting period with a duration TIMEOUT.
+<<<
+
 [/SECTION]
 
 [SECTION]
--- a/docs/zephyr_02_functional_requirements.sdoc
+++ b/docs/zephyr_02_functional_requirements.sdoc
@@ -1173,6 +1173,20 @@ the Zephyr RTOS shall accomplish all of the following:
 <<<
 
 [REQUIREMENT]
+UID: ZEP-131
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Counting Semaphore
+TITLE: Semaphore Take w/o blocking when not Available
+STATEMENT: >>>
+While a Counting Semaphore count is 0,
+
+When a thread takes that Counting Semaphore without allowing blocking,
+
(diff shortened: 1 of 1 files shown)


gregshue:
Please remove the `Where` conditionals. The 3/14/24 Focus Group decided to postpone dealing with configuration items.

docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -1317,6 +1317,248 @@ As a Zephyr OS user I want to... ???,
...
+- initial count of the Counting Semaphore; and
+- limit of the Counting Semaphore.
+<<<
+
+[REQUIREMENT]
+UID: ZEP-130
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Counting Semaphores
+TITLE: Counting Semaphores Static Object Init Failures Cause Build Errors
```
nashif, line 1363: this seems to be based on a BUILD_ASSERT that exists in K_SEM_DEFINE? Does this qualify for a dedicated requirement for each object we statically define or does this mean we will have a requirement for each BUILD_ASSERT in the code?

Maybe expand on the requirement above (ZEP-129) using EARS and say

".... WHERE the limit is larger or equal to the initial count and the limit is less the maximal limit specified by the system"

Error checking and paramter verificatio is a topic on its own, because Zephyr by design does not check every input and does does not fail gracefully (when not in userspace mode), see following comments...

docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -1317,6 +1317,248 @@ As a Zephyr OS user I want to... ???,
...
+TYPE: Functional
+COMPONENT: Counting Semaphores
+TITLE: Counting Semaphore Dynamic Object Init Failures Reports Failure Reason
+STATEMENT: >>>
+While a Counting Semaphore is dynamically allocated,
+
+If the Counting Semaphore initialization is configured with an invalid set of initialization parameter values,
+
+the Counting Semaphore initialization shall report the reason for the failure.
+<<<
```
nashif, line 1384: ditto, parameter checks should be dealy with globally and I do not think it is something we should do for each API call, which this requirement is trying to describe (k_sem_init)

docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -1317,6 +1317,248 @@ As a Zephyr OS user I want to... ???,
...
+
+Zephyr shall report the current count of the Counting Semaphore.
+<<<
+
+[REQUIREMENT]
+UID: ZEP-133
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Counting Semaphores
+TITLE: Counting Semaphore Current Count Request Invalid Parameter is Catastrophic Error
```
nashif, line 1403: another one, that is actually not true, no parameter validation happens in k_sem_count_get right now, this happens only in userspace and parameter validation is covered in 1 single requirement (ZEP-82) already, so no need to repeat this for each interface

nashif (changes requested):
few comments, did not go through the whole thing yet.

docs/zephyr_02_functional_requirements.sdoc
Timotheous: If the initialization parameters are invalid and the build completes without error it can result in unpredictable runtime behavior.  Checking the range of parameters during build is the same as verifying them when entering runtime initialization function.  Just good practice for safe coding.

There should be a global requirement that "All statically initialized objects have their parameters validated during build whenever possible."  Repeating the requirement where every it applies could create too much noise.

docs/zephyr_02_functional_requirements.sdoc
Timotheous: As long as the general requirement is clearly stated at a higher level and traces down to this requirement.  This should not be needed.

docs/zephyr_02_functional_requirements.sdoc
Timotheous: UIDs  ZEP-130, ZEP-131, ZEP-133, ZEP-135, ZEP-138, ZEP-130, ZEP-142 should all be covered by higher level global requirements to validate parameters when initializing an object.  
ZEP-82 covers parameter checking of a function call.
Don't know if there is one for parameter checking of a statically initialized object.

docs/zephyr_02_functional_requirements.sdoc
gregshue: @Timotheous is exactly right about the implicit higher-level requirement.

I've worked in an RM environment where the dictate was that requirements are complete and stand-alone. In that case there are no dependencies on document context. The advantage is downstream requirements have to link back to only one item per requirement and finding missing links becomes MUCH easier. The disadvantage is that statements propagate out. I wouldn't describe this repetition as noise. Every occurrence (and lack of occurrence) is important. Additionally, compound conditionals will exist. I don't think we can find an efficient organization for many of these compound conditions.  (I've already run across a triple-condition.)

[EARS syntax](https://alistairmavin.com/ears/) explicitly uses:
- WHERE for optional feature requirements;
- WHEN for state-driven requirements; and
- IF <trigger>, THEN for unwanted behavior requirements.

As far as I can tell this is fundamentally an unwanted behavior, secondarily a state, and certainly not an optional feature.

Yes, error checking and parameter verification is another topic that comes down from a higher layer.  Some regulations will impose policy. Some users will have very high priority needs in this area. Some behaviors will be left up to the Architecture Decisions. The most important point is unspecified behavior is implicitly insecure. IoT security regulations assessments require acceptable justifications for each non-compliant occurrence.

docs/zephyr_02_functional_requirements.sdoc
gregshue: Make sure to think about what further down architecture documents, interface specs, design docs, implementations, and tests must trace back to. Then think about the challenge of finding missing traces.

docs/zephyr_02_functional_requirements.sdoc
gregshue: I saw that it was not true. I added this to kick off discussion about capturing requirements on behavior for ALL public interfaces. The behavior is undefined when an invalid pointer (e.g., NULL) is passed in. It would make much more sense if the parameters were checked when RUNTIME_ERROR_CHECKS was enabled and disabled by NO_RUNTIME_CHECKS.

This is a case where I, as a product manufacturer, believe the code is wrong because it doesn't meet my needs.

docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -1317,6 +1317,248 @@ As a Zephyr OS user I want to... ???,
...
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Counting Semaphores
+TITLE: Counting Semaphore dynamic initialization
+STATEMENT: >>>
+Zephyr Counting Semaphores shall expose an interface to dynamically initialize a Counting Semaphore which requires the following information:
+
+- initial count of the Counting Semaphore; and
+- limit of the Counting Semaphore.
+<<<
```
Timotheous, line 1356: In a counting semaphore there are three fundamental values.  Min Count(limit), Max Count(initial count), and Timeout.  Where is Timeout defined?  
The rest of the functional requirements should be in terms of these values.

docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -1317,6 +1317,248 @@ As a Zephyr OS user I want to... ???,
...
+TYPE: Functional
+COMPONENT: Counting Semaphores
+TITLE: Counting Semaphore Take Request From Interrupt Context with Non-Immediate Timeout is Catastrophic Error
+STATEMENT: >>>
+When the Counting Semaphore Take is requested from Interrupt Context,
+
+If the Counting Semaphore Take timeout is non-immediate,
+
+Zephyr shall report a catastrophic failure.
+<<<
```
Timotheous, line 1558: Allowing a Semaphore Take in Interrupt context where Timeout is supported is inherently unsafe. It will cause a deadlock. Exchanging the deadlock for a catastrophic failure is not any safer.  And there is the issue of blocking and unblocking threads. Not all RTOSs will do that from inside an interrupt.  
Not sure how to react to this requirement except with great apprehension.

docs/zephyr_02_functional_requirements.sdoc
gregshue: @Timotheous I expect you are correct that the UIDs you listed are provided at a _higher level_ by other requirements. If we don't propagate those requirements to lower level elements we will end up with MANY links to the higher level requirements. It will probably be too many to find gaps in.

IIRC, when an architecture definition process completes:
- all the requirements on the (Element) are delegated out to each of the sub-(Element)s;
- all the interactions and requirements on the interfaces are captured and traced back to the requirements on the - (Element) to verify all functionality has been met.

Though the explicit delegation of requirements to each specific element seems "repetitious" the subject of the statements changes and those responsible for the sub-(Element) do not have to dig through a mountain of higher requirements trying to decide what does and does not apply.

docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -1317,6 +1317,248 @@ As a Zephyr OS user I want to... ???,
 Architectural & interface REQ
 <<<
```
Timotheous, line 1319: General NitPick;  
Too much Console Font gives me migraines.

The titles can be simplified. 
- Remove "Counting Semaphore" since it is already in the COMPONENT above.
- Remove verbiage that is in the STATEMENT below, except the minimum needed to make the title unique.

Some examples: 

Counting Semaphore Exposes Current Count  
Get Current Count

Counting Semaphores Exposes Ability to Reset Count to Zero
Reset Count

Counting Semaphore Give unblocks pending Take by another thread
Give with pending Take

Counting Semaphore Take when Count is > 0
Take when Count is > 0

docs/zephyr_02_functional_requirements.sdoc
gregshue: In Zephyr's design the timeout is a parameter of only the Take operation.

docs/zephyr_02_functional_requirements.sdoc
gregshue: Understood. Here is why I wrote them the way I did:

1. In general, each requirement document needs to be extensible (whether by internal or external future additions). Making the Title text unique and meaningful within an open set is just another case of namespace management.
2. As laid out now, the low-level requirements document is extremely broad. I expect someone (in addition to me) will import the entire document into another commercial tool and work with it there. I have worked with a few different ones and searches return the item titles without the section context. The "repetitive" information is for usability in that context. (NOTE: Hovering over the title in those GUIs may bring the description up in a tool tip, but sometimes the requirement statement is larger than the tool tip can display. Some CA Title 24 Building Energy Code regulation statements are well over 100 words long even after breaking them into singular statements.)
3. We must expect these requirements will be inherited as part of a down-stream product. What may be ubiquitous here may not be in the shipping product. At this point everything we are considering is still within the Kernel. To be reusable the requirements must be scoped accordingly. This will become more apparent when we start to include functionality outside the kernel (e.g., HALs, drivers, logging?, build system).

I think breaking the "low-level" requirements document into a separate document for each component (kernel object) will help greatly. Searches can be limited to the document and missing links will be far easier to identify. At that point the titles can be simplified.

docs/zephyr_02_functional_requirements.sdoc
gregshue: This is a great time to figure out what users require. ;-)

BTW, the current implementation asserts if called from interrupt context. I'm not sure what it does when asserts are disabled.

docs/zephyr_02_functional_requirements.sdoc
gregshue: BTW, the reference I'm using for the architecture process is [here](http://swebokwiki.org/Chapter_2:_Software_Design#Software_Structure_and_Architecture).

docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -1317,6 +1317,248 @@ As a Zephyr OS user I want to... ???,
...
+TITLE: Counting Semaphores
+
+[REQUIREMENT]
+UID: ZEP-127
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Counting Semaphores
+TITLE: Build-time registration for pre-kernel init phase notification
+STATEMENT: >>>
+Zephyr Counting Semaphores shall build-time register with Zephyr System Initialization for notification of running in the Zephyr Pre-Kernel initialization phase.
```
tobiaskaestner, line 1330: **issue:** This formulation introduces a lot of jargon/technical terms which as of today cannot be resolved from within the RQTs spec:
* build-time register
* Zephyr System Initialization
* Zephyr Pre-Kernel initialization phase

IMHO, We need to re-state the problem these constructs have helped us solve.

docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -1317,6 +1317,248 @@ As a Zephyr OS user I want to... ???,
...
+[REQUIREMENT]
+UID: ZEP-128
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Counting Semaphores
+TITLE: Count Semaphore pre-Kernel notification initializes statically defined semaphore objects
+STATEMENT: >>>
+When the Counting Semaphores service is notified of the Zephyr Pre-Kernel System Initialization phase,
+
+Zephyr Counting Semaphores shall initialize all statically defined semaphores.
```
tobiaskaestner, line 1342: **note:** same as my previous comment

docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -1317,6 +1317,248 @@ As a Zephyr OS user I want to... ???,
...
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Counting Semaphores
+TITLE: Counting Semaphore Current Count Request Invalid Parameter is Catastrophic Error
+STATEMENT: >>>
+When the Counting Semaphore current count is requested,
+
+If the Counting Semaphore current count parameters are invalid,
+
+Zephyr shall report a catastrophic failure.
```
tobiaskaestner, line 1409: **note:** another undefined term: catastrophic failure

docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -1317,6 +1317,248 @@ As a Zephyr OS user I want to... ???,
...
+[REQUIREMENT]
+UID: ZEP-136
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Counting Semaphores
+TITLE: Counting Semaphore Give unblocks pending Take by another thread
+STATEMENT: >>>
+While a thread is blocked waiting to Take from this Counting Semaphore,
+
+When the Counting Semaphore Give is requested,
```
tobiaskaestner, line 1450: **suggestion:** It would be easier to read, "When a running task gives the semaphore ..."

Also the statement as such is most likely inprecise. An arbitrary number of tasks can wait on a semaphore. When the semaphore is given, only the highest priority (that's where determinism kicks in) thread is unblocked.

docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -1317,6 +1317,248 @@ As a Zephyr OS user I want to... ???,
...
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Counting Semaphores
+TITLE: Counting Semaphore Take blocked when Count is 0
+STATEMENT: >>>
+While the Count of this Counting Semaphore is 0,
+
+When the Counting Semaphore Take is requested,
+
+the Counting Semaphore shall block the thread requesting a Counting Semaphore Take.
```
tobiaskaestner, line 1517: **nitpick:** "... requesting this Counting Semaphore."

docs/zephyr_02_functional_requirements.sdoc
tobiaskaestner: Greg is right though, current API docs explicitly allow to take a semaphore from ISR context under the constraint that the timeout has to be [K_NO_WAIT](https://github.com/zephyrproject-rtos/zephyr/blob/main/kernel/sem.c#L136)

docs/zephyr_02_functional_requirements.sdoc
gregshue: The additional technical terms come from decomposing higher layers of the system. Unfortunately, we started at the implementation instead of the user needs/stakeholder constraints.

docs/zephyr_02_functional_requirements.sdoc
gregshue: Agreed.  Thank you for the suggestion.

docs/zephyr_02_functional_requirements.sdoc
gregshue: Agreed. In other work I have used "said <item>". ;-) Either correction seems clear. Let's decide on one so we can be consistent.

docs/zephyr_02_functional_requirements.sdoc
nashif: of course Greg is right, this is not the only place where it happens and where ASSERTs are using to perform error checking. This has been the design of Zephyr since the first day: enable asserts during development, get your code in order, then disable asserts and ship it. We then introduced checkif macros to turn those into runtime checks and still maintain footprint under control, we also introduced userspace with its own error checking and input validation stubs. 

Bottom line, in my previous comments here I am not disagreeing to error checking and input validation, but we need figure out how to specify it effectively in the requirements without going through each error checking in the code and creating a requirement for it, this way we might actually missing some very hot spots, because the requirements are just describing the existing code logic and not the implemented concept.

nashif: https://docs.zephyrproject.org/latest/kernel/usermode/overview.html#high-level-policy-details

> For performance and footprint reasons Zephyr normally does little or no parameter error checking for kernel object or device driver APIs. Access from user mode through system calls involves an extra layer of handler functions, which are expected to rigorously validate access permissions and type of the object, check the validity of other parameters through bounds checking or other means, and verify proper read/write access to any memory buffers involved.

We have something, but here we are trying to deal with semaphores and keep the discussion focused on semaphores, we will get to the above and other requirements.

gregshue: Performance and footprint are at odds with defense-in-depth. For Zephyr the Semaphore API is a public interface. To minimize risks during usage and integration into a device all input data should be validated by default. This is a secure design principle echoed in ETSI 303 645 Provision 5.13-1.

AFAICT security trumps performance and footprint (for default behavior). I am fine with giving the Product Manufacturers the _option_ to disable parameter checking of a public API level. They are the only ones with the knowledge about the entire executable to decide when that is safe to do.

> here we are trying to deal with semaphores and keep the discussion focused on semaphores, 

Agreed. This appears because Semaphores can be acted upon from Interrupt, Supervisor, and User CPU contexts/modes and the Semaphore API is public. We must capture the required behavior. If this document is capturing requirements for the opaque Zephyr Kernel then the observable behavior may be to _exhibit_ a fatal error. If this document is capturing requirements for the opaque Zephyr Counting Semaphore within the transparent Zephyr Kernel then the observable behavior may be to _report_ a fatal error to the Zephyr Fatal Error Handler.

If you are already anticipating this will be covered at a higher layer requirement then I think we are getting ahead of ourselves. Eventually we will recognize that:

1. the purpose of tests is to verify the code-under-test fully meets the _requirements_ delegated to it,
2. the purpose of code coverage is to verify the code-under-test does nothing else,
3. there are more levels than High/Low.

I think it is wiser to capture everything for the Counting Semaphore sub-Element (e.g., kernel/sem.c, zephyr/kernel.h::k_sem*) before we move up a level to the opaque requirements on Zephyr Kernel. This pattern of moving up a decomposition level will re-occur with `arch`, `hal*`, executables, sources vs tools vs artifacts, etc.

gregshue:
Please rework this PR to reflect the latest proposal developed in collaboration with Stanislav.

PR #39, opened 2024-03-26
Reqs: Drop c library requirements

simhein:
Drop the c library requirements because they are not in the scope of kernel/* and the safety WG right now.


docs/zephyr_02_functional_requirements.sdoc
```diff
@@ -108,190 +108,6 @@ RELATIONS:
 
 [/SECTION]
 
-[SECTION]
```
stanislaw, line 111: These requirements are actually good in themselves, they are just out of scope. I feel like deleting them will lead to losing the work previously done. Also, someone will have to remember that this has been removed but used to exist, and people outside of the group will not be aware at all.

How about creating the third document called `Zephyr Requirements Backlog` where this and other out-of-scope sections would be kept? This way there will be a good visibility into the possible future scope right here in this repository.

simhein: good point. I will close this PR.

PR #56, opened 2024-04-05
Software requirements: create system init fragment

gregshue:
Summary of the changes:

- Fragment document created for a section "System Initialization".

**StrictDoc 0.0.53 version is needed for this changeset to work.**

### DIFF
--- a/docs/software_requirements/index.sdoc
+++ b/docs/software_requirements/index.sdoc
@@ -18,6 +18,9 @@ FILE: device_driver_api.sdoc
 FILE: exception_and_error_handling.sdoc
 
 [DOCUMENT_FROM_FILE]
+FILE: system_initialization.sdoc
+
+[DOCUMENT_FROM_FILE]
 FILE: file_system.sdoc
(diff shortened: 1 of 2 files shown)


simhein:
Creating empty placeholder fragments isn't a good practice in my opinion. These can be easily forgotten or a duplicated set could be created

gregshue:
> Creating empty placeholder fragments isn't a good practice in my opinion. These can be easily forgotten or a duplicated set could be created

Good point. I have added FREETEXT with a TBD so that it is no longer going to be forgotten. 

The point of this commit is to create the fragment file for system initialization requirements PRs to be developed against. If we delay the file creation until the first requirement draft then could end up with competing PRs creating different fragment filenames. We avoid that by defining the fragment file independent of any requirement statement. Please remember that RM and Design/Architecture are top-down activities. The decision _that_ there will be a System Initialization section => file must happen before the contents within that file are determined.

simhein (approved):

PR #68, opened 2024-05-28
req: Generate UIDs with new prefixes

simhein:
Generate the UIDs for the specific sections with new prefixes.
Updated the linking of the new generated system UIDs in the software requirements.


docs/software_requirements/hw_arch_interface.sdoc
```diff
@@ -1,12 +1,12 @@
 [DOCUMENT]
 TITLE: Hardware Architecture Interface
-REQ_PREFIX: ZEP-
+REQ_PREFIX: ZEP-SRS-14
```
Timotheous, line 3: Prefix should end in a dash so the Requirements are uniquely formatted.  
Maintain the pattern from previous documents.

Timotheous:
This looks good except for the missing "-" in the Hardware Architecture Interface REQ_PREFIX

This needs to be approved and merged as soon as possible to minimize the impact on requirements that are being written now.

docs/software_requirements/hw_arch_interface.sdoc
simhein: good catch thanks, fixed.

simhein:
It seems there are no objections to go on with this. I will merge it to move forward and to minimize the impact of following PRs.

PR #78, opened 2024-07-30
Req: Add file licence to all files

simhein:
Add Apache-2.0 licence to all files in the
requirements repository.


nicpappler (approved):
OK

PR #79, opened 2024-07-31
Req: Update grammar for Type field

simhein:
Update the file grammar for the Type filed to only allow the values "Functional" or "Non-Functional".


kestewart:
Restriction to the grammar makes sense (so auto checking),  so +1 on adding.
Need to get the build fixed though.

nicpappler (approved):
Makes much sense from a safety/quality and consistency perspective. +1 from me.

PR #83, opened 2025-03-12
Mailbox requirements

jkey-eng:
I've created requirements for the mailbox module in support of #63.

I used Perplexity to generate an initial list of requirements, reviewed them with the [Mailboxes documentation](https://docs.zephyrproject.org/latest/kernel/services/data_passing/mailboxes.html) and did some simple verbiage updates to be more consistent with some other requirements I've reviewed.

Looking to get some early feedback on whether this is what the team is looking for before I spend too much more time reviewing. It didn't take too long to do and seems like a good way to get the ball rolling forward.

Remaining Work:
- Comparison between source code & requirements
- Comparison between requirements & tests


docs/software_requirements/mailboxes.sdoc
```diff
@@ -0,0 +1,439 @@
...
+  VALUE: ZEP-SYRS-22
+
+[REQUIREMENT]
+UID: ZEP-SRS-22-24
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Mailboxes
+TITLE: Priority-Based Message Delivery
+STATEMENT: >>>
+When multiple threads are waiting on an empty mailbox, the Mailbox module shall deliver the next message to the highest priority thread that has waited longest.
```
tobiaskaestner, line 355:
```suggestion
When multiple threads are waiting on an empty mailbox object, the Zephyr RTOS shall deliver the next message to the highest priority thread that has waited the longest.
```

tobiaskaestner: Not sure if we need to be more precise here, ie. the threads in question all have to wait for the message from the same sender to make the rule of preference apply. That is, if X has higher priority than Y and X is waiting for a message from A while Y is waiting on a message from B, when B delivers its message first, then also Y will be scheduled to receive it first.

docs/software_requirements/mailboxes.sdoc
```diff
@@ -0,0 +1,439 @@
...
+  VALUE: ZEP-SYRS-22
+
+[REQUIREMENT]
+UID: ZEP-SRS-22-25
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Mailboxes
+TITLE: FIFO Message Delivery
+STATEMENT: >>>
+When multiple messages are available in the send queue, the Mailbox module shall deliver them in FIFO (First In, First Out) order to waiting threads.
```
tobiaskaestner, line 371: Again, this is only true IMHO for message from the same sender, ie. only messages from matching sender to matching receivers are delivered in order. As in the above case usage of K_ANY for rx/tx thread may complicate things further.

docs/software_requirements/mailboxes.sdoc
```diff
@@ -0,0 +1,439 @@
...
+  VALUE: ZEP-SYRS-22
+
+[REQUIREMENT]
+UID: ZEP-SRS-22-26
+STATUS: Draft
+TYPE: Non-Functional
+COMPONENT: Mailboxes
+TITLE: Mailbox Quantity Support
+STATEMENT: >>>
+The Mailbox module shall support an arbitrary number of mailboxes, limited only by available RAM in the system.
```
tobiaskaestner, line 384:
```suggestion
The Zephyr RTOS shall support an arbitrary number of mailbox objects, limited only by available RAM in the system.
```

docs/software_requirements/mailboxes.sdoc
```diff
@@ -0,0 +1,439 @@
...
+  VALUE: ZEP-SYRS-22
+
+[REQUIREMENT]
+UID: ZEP-SRS-22-27
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Mailboxes
+TITLE: Resource Cleanup
+STATEMENT: >>>
+The Mailbox module shall ensure proper cleanup of message resources when operations are canceled or time out to prevent resource leaks.
```
tobiaskaestner, line 400:
```suggestion
The Zephyr RTOS shall ensure proper cleanup of mailbox object resources when operations are canceled or time out to prevent resource leaks.
```

tobiaskaestner: IMHO, the problem with this kind of  requirement is in the adjective. What do we mean with 'proper'? I know that above I made the reverse point for 'synchronous' and 'asynchronous' to imply blocking and non-blocking respectively. Still, in the context of computer science they seem to be somewhat better defined than 'proper' or 'correct'

docs/software_requirements/mailboxes.sdoc
```diff
@@ -0,0 +1,439 @@
...
+  VALUE: ZEP-SYRS-22
+
+[REQUIREMENT]
+UID: ZEP-SRS-22-28
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Mailboxes
+TITLE: Error Reporting
+STATEMENT: >>>
+If an error occurs during any mailbox operation, the Mailbox module shall return an appropriate error code indicating the type of failure.
```
tobiaskaestner, line 416:
```suggestion
If an error occurs during any mailbox operation, the Zephyr RTOS shall return an appropriate error code indicating the type of failure.
```

tobiaskaestner: We mentioned timeouts at various places already. So strictly speaking this requirement does overlap with some of the others while not really saying what else might lead to an erroneous situation. Actually, to me only out-of-memory comes to mind ?

docs/software_requirements/mailboxes.sdoc
```diff
@@ -0,0 +1,439 @@
...
+  VALUE: ZEP-SYRS-22
+
+[REQUIREMENT]
+UID: ZEP-SRS-22-29
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Mailboxes
+TITLE: Invalid Parameter Handling
+STATEMENT: >>>
+The Mailbox module shall handle invalid parameters by returning appropriate error codes rather than causing system failures.
```
tobiaskaestner, line 432: Okay, that's the third option besides timeouts and out-of-memory ;-)

tobiaskaestner (changes requested):
First of all, many thanks for the PR. 
My main concern atm is to not introduce another subject for this set of requirements, ie. the Mailbox module. Where I too notice I have replaced this with the subject Zephyr RTOS as this is what we have been using in the other requirements (eg. semaphore) so far. 
The rationale behind this is to avoid irritations as to where these 'modules' need to be defined. 
Other than that, the other points relate to avoiding requirements that say what is not provided and how to deal with error cases. I'd like to see thoughts from the others on those two points, too.

docs/software_requirements/mailboxes.sdoc
jkey-eng: Split this into two requirements. One focused on the prioritization, another to handle the edge case where two threads have equal priority.

docs/software_requirements/mailboxes.sdoc
jkey-eng: This looks like another generated requirement that I should have removed. 
I don't see any functions in the API that would cancel or drive resource leaks if they were to time out.

docs/software_requirements/mailboxes.sdoc
jkey-eng: Removed.

Agreed this is redundant / unnecessary.

jkey-eng (commented):
@tobiaskaestner Thanks for the review and feedback!

I've made updates accordingly.
It doesn't look like the feedback stayed in-line, not sure if that's due to something I did on my end.

docs/software_requirements/mailboxes.sdoc
```diff
@@ -0,0 +1,439 @@
...
+  VALUE: ZEP-SYRS-22
+
+[REQUIREMENT]
+UID: ZEP-SRS-22-8
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Mailboxes
+TITLE: Buffer Data Transfer
+STATEMENT: >>>
+Where message data is in a buffer, the Mailbox module shall correctly handle the data transfer between the sending and receiving threads.
```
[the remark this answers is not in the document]
nicpappler: Commenting just for the case this one won't be removed, it definitely needs brushing up to be accepted: From a formal perspective, the term "correctly handle" is ambiguous, we need to specify what is the expected bahaviour for it to be "handled correctly".

jkey-eng:
@tobiaskaestner Have you had a chance to review the requirement updates I've made in response to your feedback?

docs/software_requirements/mailboxes.sdoc
```diff
@@ -246,13 +220,13 @@ RELATIONS:
...
 [REQUIREMENT]
-UID: ZEP-SRS-22-17
+UID: ZEP-SRS-22-15
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Mailboxes
 TITLE: Message Data Retrieval
 STATEMENT: >>>
-The Mailbox module shall provide a mechanism for a thread to retrieve mailbox message data into a buffer and complete the message processing.
+The Zephyr RTOS shall provide a mechanism for a thread to retrieve message data via a mailbox object into a buffer and complete the message processing.
```
tobiaskaestner, line 229:
```suggestion
The Zephyr RTOS shall provide a mechanism for a thread to retrieve message data via a mailbox object.
```

docs/software_requirements/mailboxes.sdoc
```diff
@@ -291,39 +265,13 @@ RELATIONS:
...
-[REQUIREMENT]
-UID: ZEP-SRS-22-22
+UID: ZEP-SRS-22-18
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Mailboxes
 TITLE: Non-Anonymous Messaging
 STATEMENT: >>>
-The Mailbox module shall handle messages non-anonymously, allowing both the sending and receiving threads to know the identity of the other thread.
+The Zephyr RTOS shall handle message exchange via a mailbox object non-anonymously, allowing both the sending and receiving threads to know the identity of the other thread.
```
tobiaskaestner, line 274: the receiving thread will always know the sending thread, but with K_ANY the sending thread cannot tell which thread received the message via a mailbox

docs/software_requirements/mailboxes.sdoc
```diff
@@ -333,26 +281,26 @@ RELATIONS:
...
 [REQUIREMENT]
-UID: ZEP-SRS-22-23
+UID: ZEP-SRS-22-19
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Mailboxes
 TITLE: Receiver Thread Queuing
 STATEMENT: >>>
-While a thread is waiting to receive a message, the Mailbox module shall keep the thread in the receive queue until a matching message arrives or the waiting period expires.
+While a thread is waiting to receive a message via a mailbox object, the Zephyr RTOS shall keep the thread in the receive queue until a matching message arrives or the waiting period expires.
```
tobiaskaestner, line 290: Consider removing this at the SRS level, we can re-introduce this at the design level since this contains implementation details.

docs/software_requirements/mailboxes.sdoc
```diff
@@ -362,74 +310,55 @@ RELATIONS:
...
 [REQUIREMENT]
-UID: ZEP-SRS-22-29
+UID: ZEP-SRS-22-24
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Mailboxes
 TITLE: Invalid Parameter Handling
 STATEMENT: >>>
-The Mailbox module shall handle invalid parameters by returning appropriate error codes rather than causing system failures.
+The Zephyr RTOS shall handle invalid parameters by returning appropriate error codes rather than causing system failures.
```
tobiaskaestner, line 361: please remove 'appropriate'

tobiaskaestner (changes requested):
@parphane and I did the review together and found a couple of RQTs that we can probably remove/merge with others and a few wording suggestions.

tobiaskaestner:
> @tobiaskaestner Have you had a chance to review the requirement updates I've made in response to your feedback?

Hi @jkey-eng , just finished another round of review ;-)

jkey-eng (commented):
@tobiaskaestner @parphane I've pushed an updated commit based on your feedback and addressed the review comments. Hopefully I caught everything, it's a little tricky to understand which review comments are stale. 

As a general practice I leave resolving Github feedback to the poster as a confirmation step that the feedback has been suitable addressed.

docs/software_requirements/mailboxes.sdoc
```diff
@@ -362,74 +310,55 @@ RELATIONS:
...
 TYPE: Functional
 COMPONENT: Mailboxes
-TITLE: Resource Cleanup
+TITLE: FIFO Message Delivery
 STATEMENT: >>>
-The Mailbox module shall ensure proper cleanup of message resources when operations are canceled or time out to prevent resource leaks.
-<<<
-USER_STORY: >>>
-As a Zephyr RTOS developer, I want the system to automatically clean up resources when operations fail or time out so that my application doesn't experience memory leaks.
+When multiple messages are available in the send queue from the same sender, the Zephyr RTOS shall deliver them in FIFO (First In, First Out) order to waiting threads.
```
woodmatthews, line 251: The linked documentation ([link](https://docs.zephyrproject.org/latest/kernel/services/data_passing/mailboxes.html#receiving-a-message)) mentioned the following:
> Receiving threads do not always receive messages in a first in, first out (FIFO) order
So I was surprised to see this requirement.

jkey-eng:
Rebased the PR based on the recent changes & StrictDoc compatibility.

@tobiaskaestner Will you be reviewing and closing out requested changes?

PR #134, opened 2026-03-07
arch: add system architecture decomposition with mailbox worked example

jkey-eng:
## Summary

Adds an architectural decomposition layer to the requirements tree, using Mailboxes as the worked example of a complete traceability chain. I chose Mailboxes because I submitted and reviewed the full requirements set (PR #83), making it a known-good baseline for demonstrating the concept.

### New artifacts

- `docs/system_architecture/system_architecture.sdoc` — System architecture for the Mailbox path (Kernel → Data Passing → Mailboxes), with a COMPONENT node tracing to the system requirement.
- `docs/system_architecture/system_architecture.sgra` — shared grammar introducing the COMPONENT element type for architectural decomposition, distinct from REQUIREMENT.
- `docs/software_architecture/mailboxes.sdoc` — Software architecture COMPONENT node mapping to existing Zephyr design documentation and API reference.
- `docs/software_architecture/software_architecture.sgra` — shared grammar with COMPONENT element.

### COMPONENT element type

Introduces COMPONENT as a new grammar element for architectural nodes. COMPONENT is semantically distinct from REQUIREMENT — it represents a decomposition node in the architecture, not a testable "shall" statement. Both architecture layers use COMPONENT nodes with Parent relations to participate in the traceability chain.

### Traceability model

Four-layer single-point traceability through the mailbox path:

```
ZEP-SYRS-25          (REQUIREMENT — system requirement)
  ↑ Parent
ZEP-SYSARCH-MBOX     (COMPONENT — system architecture)
  ↑ Parent
ZEP-SWARCH-MBOX      (COMPONENT — software architecture)
  ↑ Parent
ZEP-SRS-25-*         (REQUIREMENT — software requirements)
```

Each layer traces only to its immediate parent — no skip-level tracing.

### Software architecture role

Software architecture documents are mapping layers that point to existing Zephyr documentation rather than duplicating content. The COMPONENT node carries references to the authoritative design docs and API reference.

### Changes to existing artifacts

- Mailbox software requirements (`docs/software_requirements/mailboxes.sdoc`): re-parented from `ZEP-SYRS-25` to `ZEP-SWARCH-MBOX` for single-point traceability through both architecture layers.

## Next steps

Once the pattern is accepted, a follow-up PR would extend the system architecture to cover all kernel subsystems and establish the organizational framework for ongoing requirements work.

### DIFF
--- a/docs/software_architecture/mailboxes.sdoc
+++ b/docs/software_architecture/mailboxes.sdoc
@@ -1,0 +1,30 @@
+[DOCUMENT]
+TITLE: Mailboxes — Software Architecture
+
+[GRAMMAR]
+IMPORT_FROM_FILE: software_architecture.sgra
+
+[TEXT]
+STATEMENT: >>>
+SPDX-License-Identifier: Apache-2.0
+<<<
+
+[COMPONENT]
+UID: ZEP-SWARCH-MBOX
+STATUS: Draft
+COMPONENT: Mailboxes
+TITLE: Mailbox Software Architecture
+STATEMENT: >>>
+Software architecture for the mailbox subsystem.
+
+Design documentation:
+`Mailboxes <https://docs.zephyrproject.org/latest/kernel/services/data_passing/mailboxes.html>`_
+
+API reference:
+`Mailbox APIs <https://docs.zephyrproject.org/latest/doxygen/html/group__mailbox__apis.html>`_
+
+Kernel source: kernel/mailbox.c
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYSARCH-MBOX
--- a/docs/software_requirements/mailboxes.sdoc
+++ b/docs/software_requirements/mailboxes.sdoc
@@ -24,7 +24,7 @@ As a Zephyr RTOS developer, I want to be able to initialize a mailbox instance s
 <<<
 RELATIONS:
 - TYPE: Parent
-  VALUE: ZEP-SYRS-25
+  VALUE: ZEP-SWARCH-MBOX
 
 [REQUIREMENT]
 UID: ZEP-SRS-25-2
@@ -40,7 +40,7 @@ As a Zephyr RTOS developer, I want to be able to statically define and initializ
 <<<
 RELATIONS:
 - TYPE: Parent
-  VALUE: ZEP-SYRS-25
+  VALUE: ZEP-SWARCH-MBOX
 
 [REQUIREMENT]
 UID: ZEP-SRS-25-3
@@ -56,7 +56,7 @@ As a Zephyr RTOS developer, I want to be able to send messages of different size
 <<<
 RELATIONS:
 - TYPE: Parent
-  VALUE: ZEP-SYRS-25
+  VALUE: ZEP-SWARCH-MBOX
 
 [REQUIREMENT]
 UID: ZEP-SRS-25-4
@@ -69,7 +69,7 @@ The Zephyr RTOS shall handle the data transfer between mailbox objects of the se
 <<<
 RELATIONS:
 - TYPE: Parent
-  VALUE: ZEP-SYRS-25
+  VALUE: ZEP-SWARCH-MBOX
 
 [REQUIREMENT]
 UID: ZEP-SRS-25-5
@@ -85,7 +85,7 @@ As a Zephyr RTOS developer, I want to be able to send a message and wait for it 
 <<<
 RELATIONS:
 - TYPE: Parent
-  VALUE: ZEP-SYRS-25
+  VALUE: ZEP-SWARCH-MBOX
 
 [REQUIREMENT]
 UID: ZEP-SRS-25-6
@@ -101,7 +101,7 @@ As a Zephyr RTOS developer, I want to be able to limit how long my thread waits 
 <<<
 RELATIONS:
 - TYPE: Parent
-  VALUE: ZEP-SYRS-25
+  VALUE: ZEP-SWARCH-MBOX
 
 [REQUIREMENT]
 UID: ZEP-SRS-25-7
@@ -117,7 +117,7 @@ As a Zephyr RTOS developer, I want to be able to send a message without waiting 
 <<<
 RELATIONS:
 - TYPE: Parent
-  VALUE: ZEP-SYRS-25
+  VALUE: ZEP-SWARCH-MBOX
 
 [REQUIREMENT]
 UID: ZEP-SRS-25-8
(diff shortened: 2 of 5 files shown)


docs/software_architecture/mailboxes.sdoc
```diff
@@ -0,0 +1,30 @@
...
+[COMPONENT]
+UID: ZEP-SWARCH-MBOX
+STATUS: Draft
+COMPONENT: Mailboxes
+TITLE: Mailbox Software Architecture
+STATEMENT: >>>
+Software architecture for the mailbox subsystem.
+
+Design documentation:
+`Mailboxes <https://docs.zephyrproject.org/latest/kernel/services/data_passing/mailboxes.html>`_
```
tobiaskaestner, line 21: that highlights an interesting point to take note of: this URL is pointing to the latest version of the design doc. 

Moving forward we need a way to allow proper baselining. 
For now that's nothing that needs immediate change though.

docs/software_architecture/mailboxes.sdoc
```diff
@@ -0,0 +1,30 @@
...
+
+API reference:
+`Mailbox APIs <https://docs.zephyrproject.org/latest/doxygen/html/group__mailbox__apis.html>`_
```
tobiaskaestner, line 24: see above

docs/software_architecture/mailboxes.sdoc
```diff
@@ -0,0 +1,30 @@
...
+
+Kernel source: kernel/mailbox.c
```
tobiaskaestner, line 26: If this were the archictecture document I wonder if we should leave this detail to the detailed design specification.

in addition, if we mention the implementation file here we probably should also add a reference to the corresponding include file (which here and for all other primitives would be kernel.h - at least for now)

docs/software_requirements/mailboxes.sdoc
```diff
@@ -24,7 +24,7 @@ As a Zephyr RTOS developer, I want to be able to initialize a mailbox instance s
 <<<
 RELATIONS:
 - TYPE: Parent
-  VALUE: ZEP-SYRS-25
+  VALUE: ZEP-SWARCH-MBOX
```
tobiaskaestner, line 27: what's the reason, why you had to change the UUID?

tobiaskaestner (commented):
Many thanks for the proposal @jkey-eng .

While reading through it I can clearly see where you are coming from and I do agree that traceability is a key requirement for all documentation artifacts that we want to move forward with.

Still I wonder whether strictdoc is the right tool to capture architectural documentation and I think we should discuss this within the WG to see if we are able to reach consensus here. 

Also, if we think of System and Software architecture in terms of the V-model, i.e. like different layers at different levels of abstraction, than I'd expect the System Architecture to refer to the System Requirements as well as motivating (at least some of the) Software Requirements.

Likewise the Software Architecture would refer to both the System Architecture and the Software Requirements as its input document. 

Perhaps that's exactly what you are proposing and I just completely missed it ;-)



Your point that we want to refer to existing documentation as much as we can is very valid, yet poses other challenges when the architecture becomes somewhat split between this document and the main repository. One might argue in favor of this though, that the main documentation would thus not need any references to the requirements added. Yet perhaps, that's something desirable.

Personally, I am still more inclined to see what could be done within the existing Sphinx documentation. The sphinx documentation system is well established within the project and we might be able to get more of the relevant eyes (aka actual maintainers) onto reviewing content.

Either way, let's continue the discussion here as well as in one of the upcoming WG calls.

docs/software_architecture/mailboxes.sdoc
jkey-eng: I'll pin these to the current Zephyr release.

docs/software_architecture/mailboxes.sdoc
jkey-eng: I'll pin these to the current Zephyr release.

docs/software_architecture/mailboxes.sdoc
jkey-eng: I went back and forth on this. If I pull the source reference out, we'd need a separate design doc layer just to hold that link. That's more overhead and another document type to maintain. Keeping it in the SW architecture component means one node carries the full trace from system requirement down to source file without needing additional layers.

If we do keep it, I'll add `include/zephyr/kernel.h` alongside `kernel/mailbox.c` as you suggested. Does that work, or do you feel strongly that it belongs somewhere else?

docs/software_requirements/mailboxes.sdoc
jkey-eng: This is the core change of the PR. Previously software requirements pointed directly at the system requirement:
`ZEP-SYRS-25 ← ZEP-SRS-25-*`

Now they trace through architecture:
`ZEP-SYRS-25 ← ZEP-SYSARCH-MBOX ← ZEP-SWARCH-MBOX ← ZEP-SRS-25-*`

Each software requirement traces to the component that motivates it. The components provide the "why" between system needs and detailed requirements.

jkey-eng:
Thanks Tobias, really appreciate the thorough read. I want to make sure I'm addressing each point, so breaking them out:

**1. StrictDoc vs Sphinx**

The components here are deliberately thin. They own the traceability link, then point to the existing Zephyr docs for design content. Maintainers don't need to learn StrictDoc or change their workflow. If we move to Sphinx later, migrating is cheap.

Are you thinking sphinx-needs specifically, or something else? I didn't find it in the Zephyr tree today, so I want to make sure we're comparing like for like.

**2. V-model traceability**

I think we're saying the same thing. The re-parenting is the whole point of this PR:

```
ZEP-SYRS-25 → ZEP-SYSARCH-MBOX → ZEP-SWARCH-MBOX → ZEP-SRS-25-*
```

StrictDoc's Parent relation gives the upward link, the traceability report gives the downward view automatically.

I also evaluated linking requirements directly to other requirements (e.g., each SRS linking to both the SWARCH component and the SYRS it derives from). That created duplicate links at every level without adding traceability coverage. The component based approach gives us the same complete chain with one link per level.

Have you seen a pattern where the direct requirement to requirement linking adds value over tracing through components?

**3. Split documentation**

The core idea is using StrictDoc as our primary source of traceability so we don't create a lot of disruption to outside processes for the sake of safety compliance. The components are connectors, not content. The actual design knowledge stays in Zephyr docs where maintainers already work.

To close the awareness gap, we could add a one liner in the Zephyr RST docs linking to the generated StrictDoc page for each component. So `mailboxes.rst` would link to ZEP-SWARCH-MBOX. A maintainer editing that doc sees that safety depends on it.

**On baselining**

The broader question (how reqmgmt baselines align to Zephyr releases, drift detection, process WG intersection) is something we need to define as a WG. I'll create a tracking task for it.

Happy to discuss any of these on the next WG call or here.

PR #139, opened 2026-06-15
software: Add Stack requirements.

RICCIARDI-Adrien:
Add a set of Stack requirements based on the currently implemented API.

I'm new to the functional safety and I used the StrictDoc Web Editor to edit most of the files, so I hope I did not made too many mistakes.

### DIFF
--- a/docs/software_requirements/stacks.sdoc
+++ b/docs/software_requirements/stacks.sdoc
@@ -1,0 +1,138 @@
+[DOCUMENT]
+TITLE: Stacks
+PREFIX: ZEP-SRS-30-
+
+[GRAMMAR]
+IMPORT_FROM_FILE: software_requirements.sgra
+
+[TEXT]
+STATEMENT: >>>
+SPDX-License-Identifier: Apache-2.0
+<<<
+
+[REQUIREMENT]
+UID: ZEP-SRS-30-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stacks
+TITLE: Stack definition at compile time
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to define and initialize a stack at compile time.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-30-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stacks
+TITLE: Stack initialization at run time with an user-defined memory area
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to initialize a stack at run time, using a data buffer provided by the user.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-30-3
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stacks
+TITLE: Stack initialization at run time with automatically allocated memory area
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to initialize a stack at run time, with automatic allocation of the stack memory.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-30-4
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stacks
+TITLE: Releasing a dynamically allocated stack buffer
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to deallocate all resources previously allocated by a run-time stack definition.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-30-5
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stacks
+TITLE: Push an item to the stack
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to add a new item on top of the stack.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-26
+- TYPE: Parent
+  VALUE: ZEP-SRS-30-7
+
+[REQUIREMENT]
+UID: ZEP-SRS-30-6
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stacks
+TITLE: Pop an item from the stack
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to retrieve the current item from the top of the stack and to remove such item from the stack.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-26
+- TYPE: Parent
+  VALUE: ZEP-SRS-30-9
+
+[REQUIREMENT]
+UID: ZEP-SRS-30-7
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stacks
+TITLE: Pushing onto a full stack
+STATEMENT: >>>
+If the stack is full when an item is pushed, the Zephyr RTOS shall return an error and not add the item.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-30-8
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stacks
+TITLE: Concurrent access protection
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to push items to or pop items from the same stack object from multiple execution contexts which can run concurrently.
+<<<
+USER_STORY: >>>
+As a Zephyr RTOS user I want to access a stack from different execution contexts (like threads), including interrupt handlers.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-30-9
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stacks
+TITLE: Stack data retrieval with timeout
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to allow waiting for a stack item availability within a specified timeout.
+<<<
+USER_STORY: >>>
+As a Zephyr RTOS user I want to share a stack object between a producer and a consumer threads. The producer thread pushes items into the stack, while the consumer thread is waiting for an item to appear in the stack.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-26
--- a/docs/software_requirements/index.sdoc
+++ b/docs/software_requirements/index.sdoc
@@ -81,3 +81,6 @@ FILE: fifos.sdoc
 
 [DOCUMENT_FROM_FILE]
 FILE: mailboxes.sdoc
+
+[DOCUMENT_FROM_FILE]
+FILE: stacks.sdoc
--- a/docs/system_requirements/index.sdoc
+++ b/docs/system_requirements/index.sdoc
@@ -368,3 +368,18 @@ The Zephyr RTOS shall provide a framework to pass messages of arbitrary size bet
 <<<
 
 [[/SECTION]]
+
+[[SECTION]]
+TITLE: Stacks
+
+[REQUIREMENT]
+UID: ZEP-SYRS-26
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stacks
+TITLE: Stacks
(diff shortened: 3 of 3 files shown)


docs/software_requirements/stack.sdoc
```diff
@@ -0,0 +1,121 @@
...
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stack
+TITLE: Stack definition at compile time
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to define and initialize a stack at compile time.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-23
```
anobli: The parent id is not the right one I guess! ZEP-SYRS-23 is used for the LIFOs.
I don't see any ID for the Stack. You probably need to add one for the stack here: docs/system_requirements/index.sdoc

docs/software_requirements/stack.sdoc
```diff
@@ -0,0 +1,121 @@
...
+
+[REQUIREMENT]
+UID: ZEP-SRS-27-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stack
+TITLE: Stack definition at run time
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to define and initialize a stack at run time.
```
anobli: I am wondering if we don't need a more detailed requirement here. Basically, if we compare to ZEP-SRS-27-2, the only difference is not only run vs build time but also the arguments we give to initialize the stack (the pointer to the buffer for runtime). In my opinion, this should be reflected in the requirement.

anobli: Oh, I notice there is k_stack_init and k_stack_alloc_init and the requirement is probably for k_stack_alloc_init. Although k_stack_init doesn't seem to be really used outside of stack, this seems to be a public API so I am wondering if should not have a dedicated requirement for it.

anobli: I think we probably to state about the error handling since we are doing memory allocation at runtime here

docs/software_requirements/stacks.sdoc
```diff
@@ -0,0 +1,121 @@
...
+[REQUIREMENT]
+UID: ZEP-SRS-27-4
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stack
+TITLE: Push an item to the stack
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to add a new item on top of the stack.
+<<<
+RELATIONS:
```
anobli, line 74: We probably need a to add a relation to ZEP-SRS-27-6 since this is the push that needs to check if the stack if full or not.

docs/software_requirements/stack.sdoc
```diff
@@ -0,0 +1,121 @@
...
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stack
+TITLE: Pop an item from the stack
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to retrieve the current item from the top of the stack and to remove such item from the stack.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-23
```
anobli: I guess we probably need to add a relation to ZEP-SRS-27-8 since this we may only need to wait when we do a pop.

docs/software_requirements/stack.sdoc
```diff
@@ -0,0 +1,121 @@
...
+
+[REQUIREMENT]
+UID: ZEP-SRS-27-7
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stack
+TITLE: Stack thread safety
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to atomically access a stack items by several concurrent users.
```
anobli: This one was really confusing for me. The user story give some hint to understand it but I had to ask to an AI to explain it to me to be sure. Maybe we should rephrase it to make it less abstract.

anobli (commented):
Thank  you for this first PR! This is my first time doing this kind of review! I hope that I am doing it right and that will be useful.

docs/software_requirements/stack.sdoc
RICCIARDI-Adrien: I though that the Stack was similar enough to the already existing LIFOs concept to be able to reuse it.
I can of course create a dedicated ID if this is better.

docs/software_requirements/stack.sdoc
RICCIARDI-Adrien: You're correct, this needs to be rephrased to be understandable !

docs/software_requirements/stacks.sdoc
RICCIARDI-Adrien: That would make sense, but is a parent relation the correct semantic for this ?
Maybe we need to define new relations, like "sibling" ?

docs/software_requirements/stack.sdoc
RICCIARDI-Adrien: Do you prefer that I create a dedicated ID ?

docs/software_requirements/stack.sdoc
RICCIARDI-Adrien: I've explicitly mentioned the memory allocation mechanism for the `k_stack_alloc_init`. I looked at some other requirements that also dynamically allocate memory, but they do not seem to talk about error handling. Wouldn't it be some kind of implementation detail ?

docs/software_requirements/stack.sdoc
RICCIARDI-Adrien: That makes sense, I made the change.

docs/software_requirements/stacks.sdoc
RICCIARDI-Adrien: I used a standard parent relation.

docs/software_requirements/stack.sdoc
```diff
@@ -0,0 +1,125 @@
+[DOCUMENT]
+TITLE: Stack
```
nashif: Stacks

docs/software_requirements/stack.sdoc
nashif: Yes, please change that, it needs to point to a new section in system requirement for Stacks.

docs/software_requirements/stack.sdoc
```diff
@@ -0,0 +1,125 @@
...
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-23
+
+[REQUIREMENT]
+UID: ZEP-SRS-27-3
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stack
+TITLE: Stack deallocation at run time
```
nashif: better to use the same language used in the currentt implementation:

```Releasing a dynamically allocated stack buffer```

File: docs/software_requirements/stack.sdoc [hunk 8]
```diff
@@ -0,0 +1,125 @@
...
+  VALUE: ZEP-SYRS-23
+- TYPE: Parent
+  VALUE: ZEP-SRS-27-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-27-6
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stack
+TITLE: Stack full detection
```
nashif: Pushing onto a full stack

docs/software_requirements/stack.sdoc
```diff
@@ -0,0 +1,125 @@
+[DOCUMENT]
+TITLE: Stack
+PREFIX: ZEP-SRS-27-
```
nashif: can you use `ZEP-SRS-30-` please?

nashif (changes requested):
please address comments below and take a look at https://github.com/zephyrproject-rtos/reqmgmt/pull/143 for reference. I will remove the stack part of this PR once this is merged and align things.

docs/software_requirements/stack.sdoc
```diff
@@ -0,0 +1,125 @@
...
+  VALUE: ZEP-SYRS-23
+
+[REQUIREMENT]
+UID: ZEP-SRS-27-7
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stack
+TITLE: Concurrent access protection
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to safely access the same Stack object from multiple execution contexts which can run concurrently.
```
nashif: "safely access" is unverifiable

docs/software_requirements/stack.sdoc
```diff
@@ -0,0 +1,125 @@
...
+<<<
+USER_STORY: >>>
+As a Zephyr RTOS user I want to access a Stack from different execution contexts (like threads), including interrupt handlers. It must be granted that the operations accessing the Stack are always atomic in order to not corrupt the Stack internal structure.
```
nashif: "Safely" has no acceptance criterion, and the "it must be granted that the operations … are always atomic" is a shall-strength constraint hidden in a non-normative field — and stronger than the implementation guarantees (k_stack_pop with a non-zero timeout is not legal from an ISR).

docs/software_requirements/stack.sdoc
```diff
@@ -0,0 +1,125 @@
...
+  VALUE: ZEP-SYRS-23
+
+[REQUIREMENT]
+UID: ZEP-SRS-27-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stack
+TITLE: Stack definition at run time
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to define and initialize a stack at run time, with automatic allocation of the stack memory.
```
nashif: bundles two ideas and leaves k_stack_init without any requirement

docs/software_requirements/stack.sdoc
```diff
@@ -0,0 +1,125 @@
...
+  VALUE: ZEP-SRS-27-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-27-6
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stack
+TITLE: Stack full detection
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to detect whether the stack memory is full before adding a new item.
```
nashif: "detect whether the stack memory is full before adding a new item" describes a mechanism that doesn't exist. There is no is-full query API; the implemented contract is that k_stack_push itself returns -ENOMEM when full.
As written the requirement is unverifiable — no test can exercise a "detect before adding" mechanism. EARS unwanted-behavior form fixes it: "If the stack is full when a value is pushed, the Zephyr RTOS shall return an error and not add the value."

docs/software_requirements/stack.sdoc
RICCIARDI-Adrien: Fixed.

docs/software_requirements/stack.sdoc
RICCIARDI-Adrien: Fixed. I also changed the file name to "stacks" to better align with the other files.

docs/software_requirements/stack.sdoc
RICCIARDI-Adrien: Fixed.

docs/software_requirements/stack.sdoc
RICCIARDI-Adrien: Fixed.

docs/software_requirements/stack.sdoc
RICCIARDI-Adrien: Sure, changed.

docs/software_requirements/stack.sdoc
RICCIARDI-Adrien: OK, changed to "push items to or pop items from" to be less generic.

docs/software_requirements/stack.sdoc
RICCIARDI-Adrien: That's true, nice catch. I've created two separate requirements, one for `k_stack_init` and one for `k_stack_alloc_init`.

docs/software_requirements/stack.sdoc
RICCIARDI-Adrien: Thanks for the explanation. I removed the "it must be granted that the operations … are always atomic" sentence as it is generates more problems than it tries to solve.

docs/software_requirements/stack.sdoc
RICCIARDI-Adrien: Fixed, thanks again for the explanation. I just changed "value" by "item" because I was using the word "item" in the other requirements.

docs/software_requirements/stacks.sdoc
```diff
@@ -0,0 +1,138 @@
...
+- TYPE: Parent
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-30-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Stack
```
nashif: this still says Stack, should be Stacks

RICCIARDI-Adrien: Sorry, fixed.

nashif (approved):

Issue #119, opened 2025-11-25
Version Requirements


parphane:
Populate / extend the requirements documents with Version Requirements
- [ ] analyze existing requirements in the Zephyr [System Requirements](https://github.com/zephyrproject-rtos/reqmgmt/tree/main/docs/system_requirement) and [Software Requirements](https://github.com/zephyrproject-rtos/reqmgmt/tree/main/docs/software_requirements)
[strictDoc ](https://strictdoc.readthedocs.io/en/stable/strictdoc_01_user_guide.html) to understand whether requirements already exist for (https://docs.zephyrproject.org/apidoc/latest/group__version__apis.html) and associated source files (https://github.com/zephyrproject-rtos/zephyr/blob/main/kernel/version.c)
- [ ] if no section or file exists, add new section (https://github.com/zephyrproject-rtos/reqmgmt/tree/main/docs/system_requirements) or file (https://github.com/zephyrproject-rtos/reqmgmt/tree/main/docs/software_requirements)
- [ ] add or rework the Version Requirements
- [ ] establish linking between requirements (Software to System Requirements), using sdoc requirements linking functionality

harunobukurokawa:
I submit #140 as draft.

PR #140, opened 2026-06-24
reqmgmt: Add Kernel Version requirements (Issue #119 initial slice)

harunobukurokawa:
Add an initial narrow slice of Kernel Version requirements in support of #119.

**Changes:**
- `docs/system_requirements/index.sdoc`: Add ZEP-SYRS-26 — system-level parent for kernel version information
- `docs/software_requirements/version.sdoc`: Add ZEP-SRS-26-1 (retrieval) and ZEP-SRS-26-2 (element access)
- `docs/software_requirements/index.sdoc`: Register `version.sdoc`

**Scope note:**
This PR intentionally covers retrieval and major/minor/patch element access only.
Broader version or release policy requirements are out of scope.
The component name `Kernel Version` is used to distinguish this topic from release policy concerns;
open to renaming if the maintainers prefer `Version` or `Build Version`.

**Methodology note (first trial):**
This contribution is a trial of an AI-assisted requirements engineering workflow:
- Prior merged PRs were analyzed using AI to study contribution patterns and review tendencies in this repository.
- Requirement drafts were independently produced by two AI agents (Claude Code and Codex) and cross-reviewed between them before being submitted here.
- The workflow itself — using AI not only to write requirements but to analyze past PRs and perform multi-agent review — is offered as a sample approach for the community to evaluate.

Feedback on both the requirements and the AI-assisted process is welcome.

### DIFF
--- a/docs/software_requirements/version.sdoc
+++ b/docs/software_requirements/version.sdoc
@@ -1,0 +1,37 @@
+[DOCUMENT]
+TITLE: Kernel Version
+REQ_PREFIX: ZEP-SRS-26-
+
+[GRAMMAR]
+IMPORT_FROM_FILE: software_requirements.sgra
+
+[TEXT]
+STATEMENT: >>>
+SPDX-License-Identifier: Apache-2.0
+<<<
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Version
+TITLE: Kernel version information retrieval
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to retrieve the kernel version information for the present build.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Version
+TITLE: Kernel version element extraction
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to obtain the major, minor, and patch level elements of the kernel version information for the present build.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-26
--- a/docs/software_requirements/index.sdoc
+++ b/docs/software_requirements/index.sdoc
@@ -81,3 +81,6 @@ FILE: fifos.sdoc
 
 [DOCUMENT_FROM_FILE]
 FILE: mailboxes.sdoc
+
+[DOCUMENT_FROM_FILE]
+FILE: version.sdoc
--- a/docs/system_requirements/index.sdoc
+++ b/docs/system_requirements/index.sdoc
@@ -368,3 +368,18 @@ The Zephyr RTOS shall provide a framework to pass messages of arbitrary size bet
 <<<
 
 [[/SECTION]]
+
+[[SECTION]]
+TITLE: Kernel Version
+
+[REQUIREMENT]
+UID: ZEP-SYRS-26
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Version
+TITLE: Kernel version information
+STATEMENT: >>>
+The Zephyr RTOS shall provide kernel version information for the present build.
+<<<
+
+[[/SECTION]]


docs/software_requirements/version.sdoc
```diff
@@ -0,0 +1,37 @@
...
+<<<
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Version
+TITLE: Kernel version information retrieval
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to retrieve the kernel version information for the present build.
```
anobli, line 20: I would suggest replacing information by code which is less ambiguous about the return type (an encoded value of major, minor and patch level)

docs/software_requirements/version.sdoc
```diff
@@ -0,0 +1,37 @@
...
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Version
+TITLE: Kernel version element extraction
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to obtain the major, minor, and patch level elements of the kernel version information for the present build.
```
anobli, line 33: Same here, I would use code instead of information.

anobli: Also, I don't think we need to repeat "present build" as this already stated in system requirement.
```suggestion
The Zephyr RTOS shall provide a mechanism to obtain the major, minor, and patch level elements from the kernel version code.
```

docs/software_requirements/version.sdoc
anobli:
```suggestion
The Zephyr RTOS shall provide a mechanism to retrieve the kernel version code.
```

parphane: Could also consider simplifying to "The Zephyr RTOS shall provide a mechanism to retrieve the kernel version."

docs/software_requirements/version.sdoc
parphane: Rework and add the concept of a "compile-time mechanism" to set these (now only the obtain/get portion is covered)

Consider "The Zephyr RTOS shall provide compile-time mechanism to define the major, minor, and patch components of the kernel version."
Could also consider 1. splitting in 3 for atomicity or 2. removing the concept of major.minor.patch from the requirement to decouple from code
2. would give "The Zephyr RTOS shall provide a compile-time mechanism to define the kernel version.""

parphane (changes requested):

harunobukurokawa:
Thanks both for the detailed feedback. I've reworked the software requirements to
address all the points raised:

- Replaced the ambiguous term "information" with "kernel version code", per
  @anobli's comments.
- Removed the repeated "present build" qualifier from the software requirements,
  since it is already covered by the parent system requirement, per @anobli's
  comment.
- Added a new requirement for the compile-time mechanism that defines the kernel
  version, per @parphane's comment that the "define" side was missing (only
  obtain/get was covered).
- Split the major/minor/patchlevel extraction into three separate requirements
  (one per component) to improve atomicity, following option 1 suggested by
  @parphane.

The software requirements are now:

- ZEP-SRS-39-1: compile-time mechanism to define the kernel version
- ZEP-SRS-39-2: mechanism to retrieve the kernel version code
- ZEP-SRS-39-3: mechanism to obtain the major component from the kernel version code
- ZEP-SRS-39-4: mechanism to obtain the minor component from the kernel version code
- ZEP-SRS-39-5: mechanism to obtain the patchlevel component from the kernel version code

I also simplified the parent system requirement to leave the major/minor/patchlevel
detail to the software requirements:

> The Zephyr RTOS shall provide the kernel version for the present build.

Note: the requirement IDs moved from `ZEP-SYRS-26` / `ZEP-SRS-26-*` to `ZEP-SYRS-39`
/ `ZEP-SRS-39-*`. This is unrelated to the review — those numbers were taken by the
Atomic Service requirements merged to main in the meantime, so I rebased onto
current main and renumbered to the next free IDs.

Let me know if this breakdown looks right, or if you'd prefer a different split
(e.g., keeping the major/minor/patchlevel extraction as a single requirement).

PR #141, opened 2026-06-25
requirements: add Work Queues system and software requirements

nashif:
Add a System requirement (ZEP-SYRS-26) describing the Zephyr work queue
framework, and a new Work Queues software requirements document
(ZEP-SRS-26-*) refining it into 25 atomic, implementation-free
requirements covering work queue management, work item submission and
processing, status and cancellation, and delayable work.

The requirements reflect the currently implemented k_work API behavior
following the Route 3s approach. All software requirements link to the
ZEP-SYRS-26 system parent, and the new document is registered in the
software requirements index.

### DIFF
--- a/docs/software_requirements/work_queues.sdoc
+++ b/docs/software_requirements/work_queues.sdoc
@@ -620,6 +620,63 @@ RELATIONS:
 [[/SECTION]]
 
 [[SECTION]]
+TITLE: Triggered Work
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-51
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Triggered work item initialization
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to initialize a triggered work item at run time, associating a handler function to be executed when the work item is processed.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-27
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Submitting a triggered work item
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to submit a work item that is processed when any of a specified set of poll events becomes ready or when a specified timeout elapses, and to report the triggering outcome to the work item handler.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-52
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Resubmission of a waiting triggered work item
+STATEMENT: >>>
+If a triggered work item is submitted while it is still waiting for its poll events, then the Zephyr RTOS shall cancel the existing submission and resubmit the work item with the new set of poll events.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-28
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Cancelling a triggered work item
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to cancel a submitted triggered work item before any of its poll events becomes ready.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[[/SECTION]]
+
+[[SECTION]]
 TITLE: User Mode Work Queues
 
 [REQUIREMENT]
--- a/docs/software_requirements/work_queues.sdoc
+++ b/docs/software_requirements/work_queues.sdoc
@@ -1,0 +1,677 @@
+[DOCUMENT]
+TITLE: Work Queues
+REQ_PREFIX: ZEP-SRS-38-
+
+[GRAMMAR]
+IMPORT_FROM_FILE: software_requirements.sgra
+
+[TEXT]
+STATEMENT: >>>
+SPDX-License-Identifier: Apache-2.0
+<<<
+
+[[SECTION]]
+TITLE: Work Queue Management
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue initialization
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to initialize a work queue.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+- TYPE: Parent
+  VALUE: ZEP-SYRS-15
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-29
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue start
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to start an initialized work queue, dedicating a thread to process its work items.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+- TYPE: Parent
+  VALUE: ZEP-SYRS-15
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-41
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Processing a work queue in the calling thread
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to process the work items of an initialized work queue that has not been started, in the context of the calling thread, returning only when the work queue is stopped.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue thread priority
+STATEMENT: >>>
+When a work queue is started, the Zephyr RTOS shall allow the priority of its dedicated thread to be configured.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-30
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue thread naming
+STATEMENT: >>>
+When a work queue is started, the Zephyr RTOS shall allow a name to be assigned to the work queue's dedicated thread.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-31
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue essential thread
+STATEMENT: >>>
+When a work queue is started, the Zephyr RTOS shall allow the work queue's dedicated thread to be marked as essential to system operation.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-3
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work item processing order
+STATEMENT: >>>
+The Zephyr RTOS shall process the work items of a work queue in the order in which they were submitted.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-4
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue thread yielding
+STATEMENT: >>>
+While yielding between work items is enabled for a work queue, the Zephyr RTOS shall cause the work queue thread to yield to other ready threads between processing successive work items.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-44
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue thread yielding disabled
+STATEMENT: >>>
+While yielding between work items is disabled for a work queue, the Zephyr RTOS shall not yield the work queue thread until no work items remain queued.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-5
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue draining
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to wait until all work items submitted to a work queue have been processed.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-32
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work item submission while draining
+STATEMENT: >>>
+While a work queue is draining, the Zephyr RTOS shall reject submission of work items to that work queue unless the submission originates from the work queue's own thread.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-6
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue submission blocking on drain
+STATEMENT: >>>
+When a drain operation is initiated with submission blocking enabled, the Zephyr RTOS shall continue to prevent submission of work items to the work queue after the drain completes, until submissions are explicitly re-enabled.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-7
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue submission unblocking
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to allow work items to be submitted again to a work queue whose submissions were previously blocked.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-8
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue stop
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to stop the dedicated thread of a work queue that has been drained and is blocking new submissions.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-45
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue stop timeout
+STATEMENT: >>>
+When stopping a work queue, the Zephyr RTOS shall wait for the work queue thread to terminate for at most a caller-specified timeout.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-53
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue stop timeout expiry
+STATEMENT: >>>
+If the work queue thread has not terminated when the stop timeout expires, then the Zephyr RTOS shall return an error indicating timeout and leave the work queue running.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-46
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue stop with an essential thread
+STATEMENT: >>>
+If a stop is requested for a work queue whose dedicated thread is marked as essential to system operation, then the Zephyr RTOS shall reject the request and return an error indicating that the operation is not supported.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[[/SECTION]]
+
+[[SECTION]]
+TITLE: Work Item Submission and Processing
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-9
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work item initialization at run time
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to initialize a work item at run time, associating a handler function to be executed when the work item is processed.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-34
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work item definition at compile time
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to define and initialize a work item at compile time, associating a handler function to be executed when the work item is processed.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-37
+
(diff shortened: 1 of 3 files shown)


docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,356 @@
...
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue definition
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to define a work queue that is served by a dedicated thread.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-26
```
anobli: Maybe we should also add ZEP-SYRS-15 which defines thread.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,356 @@
...
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue thread priority
+STATEMENT: >>>
+When a work queue is started, the Zephyr RTOS shall allow the priority of its dedicated thread to be configured.
```
anobli, line 66: I feel like we would need more requirements here:
- one for describing k_work_queue_start, and maybe the fact it creates its thread
- this one, explaining behavior regarding thread priority
- probably another one, about other option given passed to thread using k_work_queue_config

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,356 @@
...
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-3
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work item processing order
+STATEMENT: >>>
+The work queue thread shall process the queued work items in the order in which they were submitted.
```
anobli: nit: inconsistency with other requirement where we have Zephyr RTOS shall

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,356 @@
...
+TITLE: Work Queue Management
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue definition
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to define a work queue that is served by a dedicated thread.
```
anobli: I am not sure about define word, it more look like init or start

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,356 @@
...
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-4
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue thread yielding
+STATEMENT: >>>
+Unless configured otherwise, the work queue thread shall yield to other ready threads between processing successive work items.
```
anobli: This doesn't respect EARS.
Maybe some thing like that would work: When yielding between work items is not disabled, the work queue ...

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,356 @@
...
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-5
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue draining
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to wait until all work items submitted to a work queue have been processed.
```
anobli, line 131: Something missing here is that we block submitting new item (except if it come from the work queue thread itself).

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,356 @@
...
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-6
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue submission blocking on drain
+STATEMENT: >>>
+When a work queue is drained, the Zephyr RTOS shall accept an option that prevents further work items from being submitted to the work queue until it is explicitly unblocked.
```
anobli: The sentence is incorrect. We give the option when we call the draining function not when the work queue has been drained.
A suggestion would be: WHEN a drain operation is initiated with submission blocking enabled, the Zephyr RTOS shall prevent further work items from being submitted to the work queue after the drain completes.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,356 @@
...
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-8
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue stop
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to stop the dedicated thread of a work queue.
```
anobli: I think we must explicitly say that only work on a drained (and refusing new item) work queue.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,356 @@
...
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-10
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work item submission to the system work queue
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to submit a work item to the system work queue.
```
anobli, line 240: We need to define what is a system work queue in my opinion

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,356 @@
...
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-13
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work item handler invocation
+STATEMENT: >>>
+When the work queue thread removes a work item from the queue, the work queue thread shall invoke that work item's handler function.
```
anobli: nit: inconsistency with other requirement where we have Zephyr RTOS shall

anobli (commented):
Here are a first set of comments. That's all I can achieve today!

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,356 @@
...
+TITLE: Work Item Submission and Processing
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-9
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work item definition
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to define a work item that associates a handler function to be executed when the work item is processed.
```
anobli: We currently have 4 different way to do it: two init functions, one for kernel mode and one for user mode. We also have two macro doing the same thing but at build time.
So, we probably need 4 requirements here.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,356 @@
...
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-15
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Waiting for a work item to complete
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to wait until a submitted work item has been processed by its work queue.
```
anobli: Has been processed is probably ambiguous. A suggestion would be `until a submitted work item's handler function has finished executing`.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,356 @@
...
+TITLE: Work Item Status and Cancellation
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-16
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work item pending status
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to determine whether a work item is currently pending.
```
anobli: Pending is ambiguous! My initial tough was queued or running but after reading the code, this seems to be much more.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,356 @@
...
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-18
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Cancellation of a running work item handler
+STATEMENT: >>>
+If a work item's handler function has already started running, then cancelling the work item shall not interrupt the execution of that handler function.
```
anobli: nit: inconsistency with other requirement where we have Zephyr RTOS shall

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,356 @@
...
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-19
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Cancellation with wait for completion
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to cancel a work item and wait until the work item is neither queued nor running.
```
anobli: `is neither queued nor running.`. Here, I would have used pending.
I said that pending is ambiguous, which is true so we need to define with a new requirement what is a pending work item, in order to using it in ZEP-SRS-26-16 and here.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,356 @@
...
+TITLE: Delayable Work
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-20
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Delayable work item definition
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to define a delayable work item that associates a handler function to be executed when the work item is processed.
```
anobli: Same here, we have both a function for doing it at runtime and a macro fr doing it a build time so we need 2 requirements in my opinion.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,356 @@
...
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-21
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Scheduling a delayable work item
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to schedule a delayable work item to be submitted to a work queue after a specified delay.
```
anobli: We need two requirements here, one for a specific work queue and one for the system queue.

nashif:
@anobli all comments addressed I hope

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,565 @@
...
+  VALUE: ZEP-SYRS-26
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-26
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work item timeout monitoring
+STATEMENT: >>>
+Where workqueue work timeout monitoring is enabled, the Zephyr RTOS shall monitor the execution duration of each work item and, if a work item handler runs longer than the configured timeout, abort the work queue thread.
```
anobli, line 510: I was not aware of that one. The requirement does the job, despite I would have probably done 2 requirement (one for the monitoring feature itself) and one about behavior in case of time out.
Aside of that, my concerns is more about the feature itself. We are just killing the queue's thread, without any notice (except a log). This looks like something we want to enable for debugging and not as a safety features as we may think when we just read the requirement. Maybe this would deserve some additional notes

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,565 @@
...
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-26
+
+[[/SECTION]]
+
+[[SECTION]]
+TITLE: Triggered Work
```
anobli, line 519: We miss init requirement here I think.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,565 @@
...
+
+[REQUIREMENT]
+UID: ZEP-SRS-26-27
```
anobli: we need to requirements here, one for the system queue and one for other queues.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,565 @@
...
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Submitting a triggered work item
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to submit a work item that is processed when any of a specified set of poll events becomes ready or when a specified timeout elapses, and to report the triggering outcome to the work item handler.
```
anobli, line 528: The requirements look quite simple regarding the documentation. As example, we miss a requirement for explaining this behavior: `Submitting a previously submitted triggered work item that is still waiting for the event cancels the existing submission and reschedules it the using the new event list` from documentation

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,565 @@
...
+The Zephyr RTOS shall provide a mechanism to cancel a submitted triggered work item before any of its poll events becomes ready.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-26
+
+[[/SECTION]]
+
+[[SECTION]]
+TITLE: User Mode Work Queues
```
anobli, line 550: nit: I realize that we don't have any system requirement about kernel and user mode.
Once we will have it (ticket is TODO), we would have to update the requirements here for parenting them with user mode.

docs/software_requirements/work_queues.sdoc
nashif: there is some ongoing work about fixing this and improving this, see https://github.com/zephyrproject-rtos/zephyr/pull/109844

kartben:
FWIW commit messages are off, e.g. they mention `ZEP-SRS-26-*` / `ZEP-SYRS-26` instead of `ZEP-SRS-38-*` / `ZEP-SYRS-37`.

nashif:
> FWIW commit messages are off, e.g. they mention `ZEP-SRS-26-*` / `ZEP-SYRS-26` instead of `ZEP-SRS-38-*` / `ZEP-SYRS-37`.

yeah, I just rebased and had to change numbering after atomic services got merged, will fix

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,573 @@
...
+TITLE: Work Queue Management
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue initialization
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to initialize a work queue that is served by a dedicated thread.
```
anobli: Do we have to talk about the thread here ? I feel like we are trying to describe some element of the work queue, unrelated to the init itself.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,573 @@
...
+  VALUE: ZEP-SYRS-15
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-41
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Processing a work queue in the calling thread
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to process the work items of a work queue in the context of the calling thread, returning only when the work queue is stopped.
```
anobli: hum, this one has confused me. Maybe we should precises that this has to be done on an initialized (and not started) work queue. As is, I initially though this was about behavior of the workqueue and not about a specific function.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,573 @@
...
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-4
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue thread yielding
+STATEMENT: >>>
+While yielding between work items is not disabled for a work queue, the Zephyr RTOS shall cause the work queue thread to yield to other ready threads between processing successive work items.
```
anobli: The double negation could be confusing. I think that `While yielding between work items is enabled` is more obvious.

anobli: Or maybe, closer to the existing code, we could say something like `While yielding between work items is disabled for a work queue, the Zephyr RTOS shall not yield the work queue thread until no work items remain queued.`

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,573 @@
...
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-8
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue stop
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to stop the dedicated thread of a work queue that has been drained and is blocking new submissions.
```
anobli, line 196: We miss expected behavior regarding the timeout (should probably be a second requirement).

anobli: Also, we need to explain behavior regarding the essential thread.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,573 @@
...
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-14
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Resubmission of an already queued work item
+STATEMENT: >>>
+If a work item is submitted while it is already queued and has not yet started running, then the Zephyr RTOS shall retain the work item at its current position in the queue.
```
anobli, line 344: We miss the case were the work is currently running while we re-submit it.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,573 @@
...
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-23
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Rescheduling a delayable work item
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to reschedule a delayable work item to a new delay, replacing any previously scheduled delay.
```
anobli: What happens if the work item has been already queued ?

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,573 @@
...
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-42
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Waiting for a delayable work item to complete
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to wait until a delayable work item's handler function has finished executing; when the work item is scheduled but not yet submitted, it shall first be submitted immediately to its work queue.
```
anobli: Here, I think we need two requirements.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,721 @@
...
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-23
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Rescheduling a delayable work item
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to reschedule a delayable work item to a new delay, replacing any previously scheduled delay.
```
anobli: Although we now have ZEP-SRS-38-49 that explains what happens if the work is already in queue or running, I think we should also explicitly say that  ZEP-SRS-38-23 applies if the work is still scheduled.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,721 @@
...
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-45
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue stop timeout
+STATEMENT: >>>
+When stopping a work queue, the Zephyr RTOS shall wait at most a caller-specified timeout for the work queue thread to terminate, and if the thread has not terminated within that period, return an error indicating timeout and leave the work queue running.
```
anobli: We need two requirements: one for the expected behavior (waiting until timeout) and one for the unexpected behavior.

nashif: split into 38-45 (bounded wait) and 38-53 (expiry: error, queue left running)

docs/software_requirements/work_queues.sdoc
nashif: now states the scheduled case explicitly; 38-49 covers queued/running.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,734 @@
...
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-23
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Rescheduling a delayable work item
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to reschedule a delayable work item to a new delay; while the work item is scheduled but not yet submitted, the previously scheduled delay is replaced by the new delay.
```
anobli: When I read the sentence, it looks like we have two different requirements so we would have to split it.
If we want to keep only one, I would suggest: `The Zephyr RTOS shall provide a mechanism to reschedule a delayable work item that is scheduled but not yet submitted, replacing the previously scheduled delay with the new delay.

anobli (commented):
Except one last comment, LGTM!

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,734 @@
...
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-4
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue thread yielding
+STATEMENT: >>>
+While yielding between work items is enabled for a work queue, the Zephyr RTOS shall cause the work queue thread to yield to other ready threads between processing successive work items.
```
tobiaskaestner, line 118: this seems particularly important to work queues that are processed in cooperative threads, as this is pointless for queues using pre-emptable threads. The requirement on thread priority does not restrict to cooperative priorities per se, so this requirement probably should do.

docs/software_requirements/work_queues.sdoc
```diff
@@ -0,0 +1,734 @@
...
+  VALUE: ZEP-SYRS-37
+
+[REQUIREMENT]
+UID: ZEP-SRS-38-44
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Work Queues
+TITLE: Work queue thread yielding disabled
+STATEMENT: >>>
+While yielding between work items is disabled for a work queue, the Zephyr RTOS shall not yield the work queue thread until no work items remain queued.
```
tobiaskaestner, line 131: See my comment on SRS-38-4

docs/software_requirements/work_queues.sdoc
tobiaskaestner: now be addressed by SRS-38-32

docs/software_requirements/work_queues.sdoc
tobiaskaestner: Now addressed by SRS-38-45

docs/software_requirements/work_queues.sdoc
nashif: https://github.com/zephyrproject-rtos/reqmgmt/compare/5abe41016e160421e1eb50e5a1cc607fda0c69b1..ea4828a77bd77eee9bf9b5d8ba18c47ff2db2fbb

docs/software_requirements/work_queues.sdoc
tobiaskaestner: now addressed by SRS-38-47

docs/software_requirements/work_queues.sdoc
tobiaskaestner: now also covered by SRS-38-49

docs/software_requirements/work_queues.sdoc
tobiaskaestner: fine for me as is right now, pending an update once the fix is available

docs/software_requirements/work_queues.sdoc
tobiaskaestner: now address by SRS-38-52

docs/software_requirements/work_queues.sdoc
tobiaskaestner: good catch, resolving nonetheless here to move forward

tobiaskaestner (approved):
As draft good for me as it stands.

Issue #24, opened 2024-01-09
Thread Requirements


simhein:
Populate / extend the requirements documents with Thread requirements.
```[tasklist]
### Tasks
```

simhein:
PR #15 Created for this Task

nicpappler:
I did the review of the requirements for Threads with focusing on our project specific requirements checklist - ok from the safety perspective, no further comments

simhein:
Thread requirements adapted, and extended.

simhein:
The Thread Reqs need adaptions to the guidelines. Issue Reopened

parphane:
Linked to PR #144

Issue #16, opened 2023-12-12
Scheduling Requirements


simhein:
Populate / extend the requirements documents with scheduling requirements

nicpappler:
- [ ] analyse existing requirements in the Zephyr High Level Requirements and the Functional Requirements sdco files regarding completeness to the source files that are associated with Scheduling (sched.c, sched_priq.h, ksched.h, timeout_q.h, idle.h)
- [ ] add the new requirements to the sections associated with Scheduling in the sdoc files
- [ ] establish linking from Functional to High Level Requirements

anobli:
Hey, could someone assign me this ticket ?
Writing requirements for scheduling won't be easy but I expect it to be very interesting!

jkey-eng:
@parphane @anobli Scheduling requirements are written and have been pending review for quite some time on #85. Feel free to use them or take over as you'd like.

anobli:
Hi Joel,

I have started doing some comments to your PR until I see your message.
I am sorry that it tooks so much time and I hope it will be faster in futur.
Do you mind if I open a new PR based on the work you have done, with the
changes I think are necessary ?
I will keep your authorship since your work was very helpful to me.

Thanks,
Alexandre

jkey-eng:
Hi Alexandre,
Feel free to take it forward, though I'd suggest continuing the PR instead of replacing it if you're going to extend the work. Appreciate the support!

PR #142, opened 2026-06-25
requirements: correct misc Threads requirements for accuracy and add new Thread requirements

nashif:
correct misc Threads requirements for accuracy and add new Thread requirements

look at individual commits for details.

### DIFF
--- a/docs/software_requirements/smp.sdoc
+++ b/docs/software_requirements/smp.sdoc
@@ -1,0 +1,56 @@
+[DOCUMENT]
+TITLE: Multicore and SMP
+REQ_PREFIX: ZEP-SRS-34-
+
+[GRAMMAR]
+IMPORT_FROM_FILE: software_requirements.sgra
+
+[TEXT]
+STATEMENT: >>>
+SPDX-License-Identifier: Apache-2.0
+<<<
+
+[REQUIREMENT]
+UID: ZEP-SRS-34-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Multi Core
+TITLE: Support operation on more than one CPU
+STATEMENT: >>>
+The Zephyr RTOS shall support operation on more than one physical CPU sharing the same kernel state.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-11
+
+[REQUIREMENT]
+UID: ZEP-SRS-34-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Multi Core
+TITLE: Running threads on specific CPUs
+STATEMENT: >>>
+The Zephyr RTOS shall provide an interface to specify the set of CPUs on which a thread is eligible to execute.
+<<<
+USER_STORY: >>>
+As a Zephyr RTOS user I want Zephyr OS to be able to specify the CPU core or the set of CPU cores on which a thread shall be executed.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-12
+
+[REQUIREMENT]
+UID: ZEP-SRS-34-3
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Multi Core
+TITLE: Exclusion between physical CPUs
+STATEMENT: >>>
+The Zephyr RTOS shall provide an interface for mutual exclusion between multiple physical CPUs.
+<<<
+USER_STORY: >>>
+As a Zephyr RTOS user I want Zephyr OS to provide synchronization mechanisms between the CPU cores and the access to common resources.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-11
--- a/docs/software_requirements/smp.sdoc
+++ b/docs/software_requirements/smp.sdoc
@@ -40,6 +40,19 @@ RELATIONS:
   VALUE: ZEP-SYRS-12
 
 [REQUIREMENT]
+UID: ZEP-SRS-34-21
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Multi Core
+TITLE: Default CPU eligibility
+STATEMENT: >>>
+When no CPU set has been specified for a thread, the Zephyr RTOS shall consider the thread eligible to execute on any CPU.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-12
+
+[REQUIREMENT]
 UID: ZEP-SRS-34-3
 STATUS: Draft
 TYPE: Functional
--- a/docs/software_requirements/thread_scheduling.sdoc
+++ b/docs/software_requirements/thread_scheduling.sdoc
@@ -46,6 +46,19 @@ RELATIONS:
   VALUE: ZEP-SYRS-17
 
 [REQUIREMENT]
+UID: ZEP-SRS-2-24
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Deadline scheduling tie-breaking
+STATEMENT: >>>
+Where earliest-deadline-first scheduling is enabled, when two or more ready threads have equal static priority and equal deadlines, the Zephyr RTOS shall treat them as threads of equal priority.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-17
+
+[REQUIREMENT]
 UID: ZEP-SRS-2-8
 STATUS: Draft
 TYPE: Functional
@@ -108,5 +121,234 @@ As a Zephyr RTOS user, I want to be able to configure my RTOS in the way, that t
 RELATIONS:
 - TYPE: Parent
   VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-12
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Selecting the current thread
+STATEMENT: >>>
+When a scheduling decision is made on a CPU, the Zephyr RTOS shall select as the current thread the highest-priority ready thread that is eligible to execute on that CPU.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-16
+- TYPE: Parent
+  VALUE: ZEP-SYRS-12
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-13
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Scheduling threads of equal priority
+STATEMENT: >>>
+When two or more ready threads share the same priority, the Zephyr RTOS shall select among them the thread that has been ready the longest.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-14
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Cooperative threads
+STATEMENT: >>>
+The Zephyr RTOS shall support cooperative threads that, once running, are not subject to preemption by other threads and remain the current thread until they yield or block.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-17
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-15
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Yielding the CPU
+STATEMENT: >>>
+The Zephyr RTOS shall provide an interface for the current thread to voluntarily relinquish the CPU to other ready threads of the same or higher priority. If no ready thread of the same or higher priority exists, the current thread shall continue to execute.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-16
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Locking the scheduler
+STATEMENT: >>>
+The Zephyr RTOS shall provide interfaces for a preemptible thread to disable preemption of itself by other threads and to subsequently restore its preemptible status.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-25
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Blocking with the scheduler locked
+STATEMENT: >>>
+While a thread has disabled preemption, when that thread becomes unready, the Zephyr RTOS shall allow other threads to execute.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-26
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Preemption status across blocking
+STATEMENT: >>>
+When a thread that has disabled preemption resumes execution after having been unready, the Zephyr RTOS shall continue to treat the thread as non-preemptible until it re-enables preemption.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-17
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Setting a thread deadline
+STATEMENT: >>>
+Where earliest-deadline-first scheduling is enabled, the Zephyr RTOS shall provide an interface to set a thread's deadline.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-17
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-18
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Per-thread time slice
+STATEMENT: >>>
+Where per-thread time slicing is enabled, the Zephyr RTOS shall provide an interface to set an individual time slice duration and expiry callback for a specific thread.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-19
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Forcing a scheduling decision
+STATEMENT: >>>
+The Zephyr RTOS shall provide an interface to request an immediate re-evaluation of the scheduling decision on the current CPU in accordance with the scheduling rules.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-27
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Forcing a scheduling decision from an interrupt
+STATEMENT: >>>
+When a re-evaluation of the scheduling decision is requested from an interrupt service routine, the Zephyr RTOS shall perform the re-evaluation upon exit from the interrupt.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-20
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Querying thread preemptibility
+STATEMENT: >>>
+The Zephyr RTOS shall provide an interface to determine whether the currently executing context can be preempted by another thread.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-21
+STATUS: Draft
+TYPE: Non-Functional
+COMPONENT: Thread Scheduling
+TITLE: Selecting the ready queue implementation
+STATEMENT: >>>
+The Zephyr RTOS shall support selecting the scheduler's ready queue implementation at build time.
+<<<
+USER_STORY: >>>
+As a Zephyr RTOS user, I want to choose a ready queue implementation that matches my system's memory, code size, and ready-thread-count profile, so that scheduling overhead fits my application.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-29
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Deferred interrupt processing thread priorities
+STATEMENT: >>>
+Where deferred interrupt processing threads are configured, the Zephyr RTOS shall reserve the highest thread priorities for them.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-17
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-22
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Preemption by deferred interrupt processing threads
+STATEMENT: >>>
+Where deferred interrupt processing threads are configured, when a deferred interrupt processing thread becomes ready, the Zephyr RTOS shall preempt the current thread, even if it is cooperative or has disabled preemption, and select the ready thread as the current thread, when either of the following conditions is satisfied:
+
+- the current thread is not a deferred interrupt processing thread
+- the current thread is a deferred interrupt processing thread of lower priority
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-17
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-28
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Deferred interrupt processing thread interruptibility
+STATEMENT: >>>
+Where deferred interrupt processing threads are configured, a running deferred interrupt processing thread shall remain interruptible by interrupt service routines.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-17
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-23
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Time slice priority threshold
+STATEMENT: >>>
+Where time slicing is enabled, the Zephyr RTOS shall exempt threads whose priority is higher than a configurable priority threshold from time slicing.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SRS-2-11
 
 [[/SECTION]]
--- a/docs/software_requirements/thread_scheduling.sdoc
+++ b/docs/software_requirements/thread_scheduling.sdoc
@@ -11,56 +11,6 @@ SPDX-License-Identifier: Apache-2.0
 <<<
 
 [[SECTION]]
-TITLE: Multi Core and SMP
-
-[REQUIREMENT]
-UID: ZEP-SRS-2-1
-STATUS: Draft
-TYPE: Functional
-COMPONENT: Multi Core
-TITLE: Support operation on more than one CPU
-STATEMENT: >>>
-The Zephyr RTOS shall support operation on more than one physical CPU sharing the same kernel state.
-<<<
-RELATIONS:
-- TYPE: Parent
-  VALUE: ZEP-SYRS-11
-
-[REQUIREMENT]
-UID: ZEP-SRS-2-2
-STATUS: Draft
-TYPE: Functional
-COMPONENT: Multi Core
-TITLE: Running threads on specific CPUs
-STATEMENT: >>>
-The Zephyr RTOS shall provide an interface to specify the set of CPUs on which a thread is eligible to execute.
-<<<
-USER_STORY: >>>
-As a Zephyr RTOS user I want Zephyr OS to be able to specify the CPU core or the set of CPU cores on which a thread shall be executed.
-<<<
-RELATIONS:
-- TYPE: Parent
-  VALUE: ZEP-SYRS-12
-
-[REQUIREMENT]
-UID: ZEP-SRS-2-3
-STATUS: Draft
-TYPE: Functional
-COMPONENT: Multi Core
-TITLE: Exclusion between physical CPUs
-STATEMENT: >>>
-The Zephyr RTOS shall provide an interface for mutual exclusion between multiple physical CPUs.
-<<<
-USER_STORY: >>>
-As a Zephyr RTOS user I want Zephyr OS to provide synchronization mechanisms between the CPU cores and the access to common resources.
-<<<
-RELATIONS:
-- TYPE: Parent
-  VALUE: ZEP-SYRS-11
-
-[[/SECTION]]
-
-[[SECTION]]
 TITLE: Thread Scheduling
 
 [REQUIREMENT]
--- a/docs/software_requirements/thread_scheduling.sdoc
+++ b/docs/software_requirements/thread_scheduling.sdoc
@@ -11,7 +11,7 @@ SPDX-License-Identifier: Apache-2.0
 <<<
 
 [[SECTION]]
-TITLE: Thread Scheduling
+TITLE: Multi Core and SMP
 
 [REQUIREMENT]
 UID: ZEP-SRS-2-1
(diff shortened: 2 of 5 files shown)


anobli (commented):
Updated and added requirements for threads look good to me.
Regarding the changes for thread scheduling, there is https://github.com/zephyrproject-rtos/reqmgmt/pull/85/ what would deserve another round of review and to be merge before.

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,255 @@ SPDX-License-Identifier: Apache-2.0
...
+UID: ZEP-SRS-2-4
 STATUS: Draft
 TYPE: Functional
-COMPONENT: Multi Core
-TITLE: Support operation on more than one CPU
+COMPONENT: Thread Scheduling
+TITLE: Scheduling a thread based on an event
 STATEMENT: >>>
-The Zephyr RTOS shall support operation on more than one physical CPU sharing the same kernel state.
+The Zephyr RTOS shall provide an interface to schedule a thread based on an event.
```
anobli: I think that event here is ambiguous! Same for scheduling ? Does that mean adding it to the queue, running it ?
The use story give some clues about but we should clarify the requirement.

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,255 @@ SPDX-License-Identifier: Apache-2.0
...
+UID: ZEP-SRS-2-5
 STATUS: Draft
 TYPE: Functional
-COMPONENT: Multi Core
-TITLE: Running threads on specific CPUs
+COMPONENT: Thread Scheduling
+TITLE: Deadline Scheduling Priorities
 STATEMENT: >>>
-The Zephyr RTOS shall provide an interface for running threads on specific sets of CPUs ( default is 1 CPU).
+Where earliest-deadline-first scheduling is enabled, the Zephyr RTOS shall, among ready threads of equal static priority, give precedence to the thread with the earlier deadline.
```
anobli: Although i perfectly understand the requirement, I don't like `give precedence`. Also, I think we are not fully respecting EARS here. I would suggest:
Where earliest-deadline-first scheduling is enabled, when two or more ready threads have equal static priority, the Zephyr RTOS shall consider the thread with the earliest deadline to have the higher priority.

anobli: Oh, and alone, I guess this is incomplete. What happen if we have thread with a deadline and another one without deadline ? Same in case where 2 threads has the same deadline (if it could happen).
We probably need additional requirements to specify it.

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,255 @@ SPDX-License-Identifier: Apache-2.0
...
+UID: ZEP-SRS-2-8
 STATUS: Draft
 TYPE: Functional
-COMPONENT: Multi Core
-TITLE: Exclusion between physical CPUs
+COMPONENT: Thread Scheduling
+TITLE: Organize running threads into a fixed list
 STATEMENT: >>>
-The Zephyr RTOS shall provide an interface for mutual exclusion between multiple physical CPUs.
+The Zephyr RTOS shall organize running threads into a fixed list of numeric priorities.
```
anobli: This looks like an implementation detail here (how instead of what).

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,255 @@ SPDX-License-Identifier: Apache-2.0
...
-UID: ZEP-SRS-2-4
+UID: ZEP-SRS-2-9
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Thread Scheduling
-TITLE: Scheduling a thread based on an event
+TITLE: Preemption support
 STATEMENT: >>>
-The Zephyr RTOS shall provide an interface to schedule a thread based on an event.
+The Zephyr RTOS shall support preemption of a running thread by a higher priority thread.
```
anobli: This is quite ambiguous. As this is stated, it looks a bit contradictory with the next one about cooperative thread. Also, what we expected is incomplete for me. When the preemption should happen (at anytime, or at some specific points). Also, by a higher priority thread is also ambiguous: the higher priority thread must be ready to preempt a running thread.

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,255 @@ SPDX-License-Identifier: Apache-2.0
...
-UID: ZEP-SRS-2-6
+UID: ZEP-SRS-2-11
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Thread Scheduling
-TITLE: Work Queue utility capable of running preemptible work items
+TITLE: Time sharing of CPU resources
 STATEMENT: >>>
-The Zephyr RTOS shall provide a thread-pooled work queue utility capable of running preemptible work items with specific scheduler priorities.
+Where time slicing is enabled, the Zephyr RTOS shall cause the current thread to relinquish the CPU after not more than a configurable time slice period whenever one or more other ready threads of the same priority exist.
```
anobli: Correct me if I am wrong, but, this only apply to preemptive thread, right? If so, we have to specify it.

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,255 @@ SPDX-License-Identifier: Apache-2.0
...
+  VALUE: ZEP-SYRS-17
 
 [REQUIREMENT]
-UID: ZEP-SRS-2-5
+UID: ZEP-SRS-2-10
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Thread Scheduling
-TITLE: Deadline Scheduling Priorities
+TITLE: Un-preemptible thread priorities
```
anobli: Nothing really wrong here but everywhere in Zephyr RTOS documentation we use cooperative thread. Maybe we should keep using it here for more clarity.

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,255 @@ SPDX-License-Identifier: Apache-2.0
...
 STATEMENT: >>>
-The Zephyr RTOS shall organize running threads by earliest deadline first priority.
+The Zephyr RTOS shall support thread priorities which cannot be preempted by other user threads.
```
anobli: I think we should remove user here. user, in the current context may be interpreted as user move (vs kernel mode).

anobli: Actually, I have noticed there is also ZEP-SRS-2-14 so I am wondering what is the real purpose of this requirement. States about priority itself (a not about the behavior of a coop thread)?

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,255 @@ SPDX-License-Identifier: Apache-2.0
...
-UID: ZEP-SRS-2-7
+UID: ZEP-SRS-2-12
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Thread Scheduling
-TITLE: Run user supplied functions in-order in a separate thread(s)
+TITLE: Selecting the current thread
 STATEMENT: >>>
-The Zephyr RTOS shall provide an interface for running user-supplied functions.
+The Zephyr RTOS shall select the highest-priority ready thread to be the current thread.
```
anobli: I think we should also mention cpu affinity (we could only select the highest priority thread that could run on the current cpu).

anobli: Actually, I don't feel comfortable with that requirement. We have some ambiguity if we consider that we have preemptive and cooperative threads. This could be read as we always run / update current with the highest ready thread which is not true in case of cooperative thread.

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,255 @@ SPDX-License-Identifier: Apache-2.0
...
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-13
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Scheduling threads of equal priority
+STATEMENT: >>>
+When several ready threads share the highest priority, the Zephyr RTOS shall run the thread that became ready earliest until it yields, blocks, or is preempted.
```
anobli: I am confused by `until it yields, blocks, or is preempted`. In my opinion, we should remove it.

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,255 @@ SPDX-License-Identifier: Apache-2.0
...
-UID: ZEP-SRS-2-8
+UID: ZEP-SRS-2-14
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Thread Scheduling
-TITLE: Organize running threads into a fixed list
+TITLE: Cooperative threads
 STATEMENT: >>>
-The Zephyr RTOS shall organize running threads into a fixed list of numeric priorities.
+The Zephyr RTOS shall support cooperative threads that, once scheduled, remain the current thread until they yield or block.
```
anobli: I think that running is probably more accurate than scheduled.

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,255 @@ SPDX-License-Identifier: Apache-2.0
...
+  VALUE: ZEP-SYRS-17
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-15
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Yielding the CPU
+STATEMENT: >>>
+The Zephyr RTOS shall provide an interface for the current thread to voluntarily relinquish the CPU to other ready threads of the same or higher priority.
```
anobli: Also, i am not sure about but should we not say what happens when there is no equal or higher priority ready thread ?
It sounds obvious but, in that case the current thread is supposed to keep running without any context switch so it might be important to mention it.

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,255 @@ SPDX-License-Identifier: Apache-2.0
...
-UID: ZEP-SRS-2-9
+UID: ZEP-SRS-2-16
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Thread Scheduling
-TITLE: Preemption support
+TITLE: Locking the scheduler
 STATEMENT: >>>
-The Zephyr RTOS shall support preemption of a running thread by a higher priority thread.
+The Zephyr RTOS shall provide an interface for a preemptible thread to temporarily prevent itself from being preempted by other threads.
```
anobli: Ambiguous / incomplete i my opinion. What happen if the thread when the thread become unready? What does temporally mean. A timeout, or do we have another mechanism for re-enabling the preemption for that thread ?
I think we need additional requirements here.

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,255 @@ SPDX-License-Identifier: Apache-2.0
...
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-19
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Forcing a scheduling decision
+STATEMENT: >>>
+The Zephyr RTOS shall provide an interface to force a scheduling decision on the current CPU so that the next thread to execute is determined immediately.
```
anobli: This looks like a requirement for k_reschedule. If this is the case, this is incomplete an confusing.
k_reschedule seems much more complex than that and if this is the case, we need to complete probably additional requirements.

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,255 @@ SPDX-License-Identifier: Apache-2.0
...
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-21
+STATUS: Draft
+TYPE: Non-Functional
+COMPONENT: Thread Scheduling
+TITLE: Selecting the ready queue implementation
+STATEMENT: >>>
+The Zephyr RTOS shall support selecting the scheduler's ready queue implementation at build time.
```
anobli: What do we expect from the different implementation? Without defining how the different ready queue are implemented, I think we should at least explain what is expected (less memory usage, or more scalable, etc)

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,255 @@ SPDX-License-Identifier: Apache-2.0
...
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-22
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Meta-IRQ thread scheduling
+STATEMENT: >>>
+Where meta-IRQ priorities are configured, the Zephyr RTOS shall schedule a ready meta-IRQ thread in preference to all other threads, including cooperative threads.
```
anobli: Correct me if I am wrong but meta-IRQ actually doing preemption ?
I think this requirement should be similar to the one about preemptive thread. Maybe it should also explicit (as this is done in documentation) that meta-irq is a thread and can also preempted by an ISR or by an higher priority meta-IRQ

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,255 @@ SPDX-License-Identifier: Apache-2.0
...
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Time slice priority threshold
+STATEMENT: >>>
+Where time slicing is enabled, the Zephyr RTOS shall exempt threads whose priority is higher than a configurable priority threshold from time slicing.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-17
```
anobli: Should have ZEP-SRS-2-11as parent.

anobli (commented):
Here are the comments regarding the thread scheduling.
Something missing and that could bring some ambiguity is the definition in priority, especially since  the scheduler consider that lower is the priority number and higher is the thread priority. 
Also, I have not seen any requirement for changing thread state to sleep and for waking it up, and busy waiting.

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,281 @@ SPDX-License-Identifier: Apache-2.0
...
+UID: ZEP-SRS-2-4
 STATUS: Draft
 TYPE: Functional
-COMPONENT: Multi Core
-TITLE: Support operation on more than one CPU
+COMPONENT: Thread Scheduling
+TITLE: Scheduling a thread based on an event
 STATEMENT: >>>
-The Zephyr RTOS shall support operation on more than one physical CPU sharing the same kernel state.
+When an event on which a thread is waiting occurs, the Zephyr RTOS shall make the waiting thread ready, so that it becomes eligible for selection by the scheduler.
```
anobli: I think we could remove `so that it becomes eligible for selection by the scheduler.`
There are other requirements defining what to do with a ready thread.

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,281 @@ SPDX-License-Identifier: Apache-2.0
...
-UID: ZEP-SRS-2-5
+UID: ZEP-SRS-2-10
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Thread Scheduling
-TITLE: Deadline Scheduling Priorities
+TITLE: Cooperative thread priorities
 STATEMENT: >>>
-The Zephyr RTOS shall organize running threads by earliest deadline first priority.
+The Zephyr RTOS shall support designating a range of thread priorities as cooperative, such that a thread executing at one of these priorities is not subject to preemption by other threads.
```
anobli: This is one overlap a little bit with ZEP-SRS-2-14. And I feel like we miss some consistency regarding preemptive and meta-IRQ (we only define priorities for cooperative threads).
Maybe we could update this requirements to describes that we have three priorities ranges for coop, preempt and meta-irq threads and complete ZEP-SRS-2-14 (with statement about coop thread being um-preemptible )

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,281 @@ SPDX-License-Identifier: Apache-2.0
...
+TITLE: Selecting the current thread
 STATEMENT: >>>
-The Zephyr RTOS shall provide an interface for running user-supplied functions.
+When a scheduling decision is made on a CPU, the Zephyr RTOS shall select as the current thread the highest-priority ready thread that is eligible to execute on that CPU.
 <<<
-USER_STORY: >>>
-As a Zephyr RTOS user, I want to be able to run functions in-order in a separated thread/threads.
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-16
```
anobli, line 136: I think we could also add ZEP-SYRS-12 which is related to cpu affinity and would explain (when it will be done) what could make a thread  eligible to run on a specific cpu.

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,281 @@ SPDX-License-Identifier: Apache-2.0
...
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-25
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Blocking with the scheduler locked
+STATEMENT: >>>
+While a thread has disabled preemption, when that thread becomes unready, the Zephyr RTOS shall allow other threads to execute and shall maintain the thread's non-preemptible status when it becomes the current thread again.
```
anobli: I think we should split it in two requirements

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,281 @@ SPDX-License-Identifier: Apache-2.0
...
-UID: ZEP-SRS-2-11
+UID: ZEP-SRS-2-19
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Thread Scheduling
-TITLE: Time sharing of CPU resources
+TITLE: Forcing a scheduling decision
 STATEMENT: >>>
-The Zephyr RTOS shall support time sharing of CPU resources among threads of the same priority.
+The Zephyr RTOS shall provide an interface to request an immediate re-evaluation of the scheduling decision on the current CPU in accordance with the scheduling rules. When invoked from an interrupt service routine, the re-evaluation shall occur upon exit from the interrupt.
```
anobli: We need to split it in two requirements here

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,281 @@ SPDX-License-Identifier: Apache-2.0
...
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-22
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Meta-IRQ thread scheduling
+STATEMENT: >>>
+Where meta-IRQ priorities are configured, when a meta-IRQ thread becomes ready, the Zephyr RTOS shall preempt the current thread, including a cooperative or scheduler-locked thread or a lower-priority meta-IRQ thread, and select the meta-IRQ thread as the current thread. A running meta-IRQ thread shall remain interruptible by interrupt service routines.
```
anobli: This should be split in two requirements

anobli (commented):
Only few remaining changes to do after I think it should be good.

nashif:
Revisited this once more: the three-range version leaked a Zephyr-coined identifier (meta-IRQ) into the requirements, which we've been deliberately avoiding elsewhere. 
Meta-IRQ is our implementation of a general concept — threads that extend interrupt handling into a schedulable context, whose readiness overrides normal preemption. 
So 2-10 now partitions the priority space into the two textbook ranges (cooperative, preemptible), and the reserved class is specified separately in implementation-independent terms: ZEP-SRS-2-29 (highest priorities reserved for deferred interrupt processing threads), 2-22 (their preemption behavior, incl. cooperative/preemption-disabled threads), 2-28 (remain ISR-interruptible).

docs/system_requirements/index.sdoc
```diff
@@ -249,9 +246,6 @@ TITLE: Thread management
 STATEMENT: >>>
 The Zephyr RTOS shall provide a framework for managing multiple threads of execution.
```
parphane: 1. Can't we merge the 2 requirements into 1 to simplify?
It is not clear (starting review), how requirements will choose to link to one or another.
2. Also consider adding the concept of SW (vs HW) thread.

=> "The Zephyr RTOS shall provide a framework for managing software threads of execution."

docs/software_requirements/threads.sdoc
parphane, line 2: Reviewed against:
1. https://docs.zephyrproject.org/latest/doxygen/html/group__thread__apis.html

**Remarks**  

- 27, 28, 29, 30, 34 cannot be linked
- Missing requirements for any function without an allocated ID

**Other design constraints**  

- 8, 15, 22, 23, 31, 32

**Data Structures** 

- 7, 12 - struct k_thread

**Macros**  

- #define K_PRIO_COOP(x)
- #define K_PRIO_PREEMPT(x)
- #define k_thread_access_grant(thread, ...)
- 35, 9, 10, 12 - #define K_THREAD_DEFINE(name, stack_size, entry, p1, p2, p3, prio, options,   delay)
- 35, 9, 10, 12 - #define K_KERNEL_THREAD_DEFINE(name, stack_size, entry, p1, p2, p3, prio,   options, delay)

**Functions**

- static void k_thread_runtime_stats_longest_frame_reset (__maybe_unused   struct k_thread *thread)
- 26 - void k_thread_foreach (k_thread_user_cb_t user_cb, void *user_data)
- void k_thread_foreach_filter_by_cpu (unsigned int   cpu, k_thread_user_cb_t user_cb, void *user_data)
- void k_thread_foreach_unlocked (k_thread_user_cb_t user_cb, void   *user_data)
- void k_thread_foreach_unlocked_filter_by_cpu (unsigned int   cpu, k_thread_user_cb_t user_cb, void *user_data)
- 24 - k_thread_stack_t * k_thread_stack_alloc (size_t size, int flags)
- 25 - int k_thread_stack_free (k_thread_stack_t *stack)
- 1, 5, 9, 10, 12 - k_tid_t k_thread_create (struct k_thread *new_thread, k_thread_stack_t *stack,   size_t stack_size, k_thread_entry_t entry, void *p1, void *p2, void   *p3, int prio, uint32_t options, k_timeout_t delay)
- FUNC_NORETURN void k_thread_user_mode_enter (k_thread_entry_t entry, void *p1,   void *p2, void *p3)
- static void k_thread_heap_assign (struct k_thread *thread,   struct k_heap *heap)
- 20 - int k_thread_join (struct k_thread *thread, k_timeout_t timeout)
- int32_t k_sleep (k_timeout_t timeout)
- static int32_t k_msleep (int32_t ms)
- int32_t k_usleep (int32_t us)
- void k_busy_wait (uint32_t usec_to_wait)
- bool k_can_yield (void)
- void k_yield (void)
- void k_wakeup (k_tid_t thread)
- __attribute_const__ k_tid_t k_sched_current_thread_query (void)
- static bool k_is_pre_kernel (void)
- 19 - static   __attribute_const__ k_tid_t k_current_get (void)
- 6 - void k_thread_abort (k_tid_t thread)
- k_ticks_t k_thread_timeout_expires_ticks (const   struct k_thread *thread)
- 33 - k_ticks_t k_thread_timeout_remaining_ticks (const   struct k_thread *thread)
- 16 - int k_thread_priority_get (k_tid_t thread)
- 2 - void k_thread_priority_set (k_tid_t thread, int prio)
- void k_thread_deadline_set (k_tid_t thread, int deadline)
- void k_thread_absolute_deadline_set (k_tid_t thread, int deadline)
- void k_reschedule (void)
- int k_thread_cpu_mask_clear (k_tid_t thread)
- int k_thread_cpu_mask_enable_all (k_tid_t thread)
- int k_thread_cpu_mask_enable (k_tid_t thread, int cpu)
- int k_thread_cpu_mask_disable (k_tid_t thread, int cpu)
- int k_thread_cpu_pin (k_tid_t thread, int cpu)
- 3 - void k_thread_suspend (k_tid_t thread)
- 4 - void k_thread_resume (k_tid_t thread)
- 14 - static void k_thread_start (k_tid_t thread)
- void k_sched_time_slice_set (int32_t slice, int prio)
- void k_thread_time_slice_set (struct k_thread *th, int32_t slice_ticks, k_thread_timeslice_fn_t expired,   void *data)
- void k_sched_lock (void)
- void k_sched_unlock (void)
- void k_thread_custom_data_set (void *value)
- void * k_thread_custom_data_get (void)
- 17 - int k_thread_name_set (k_tid_t thread, const char *str)
- 18 - const char * k_thread_name_get (k_tid_t thread)
- int k_thread_name_copy (k_tid_t thread, char *buf, size_t size)
- 21 - const char * k_thread_state_str (k_tid_t thread_id, char *buf, size_t   buf_size)

docs/software_requirements/threads.sdoc
```diff
@@ -150,29 +157,43 @@ RELATIONS:
...
-UID: ZEP-SRS-1-10
+UID: ZEP-SRS-1-11
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Threads
-TITLE: Scheduling multiple threads
+TITLE: Thread Options
 STATEMENT: >>>
-The Zephyr RTOS shall provide an interface to schedule multiple threads.
+The Zephyr RTOS shall support a set of thread options.
```
parphane: Consider adding "When creating a thread, [...]" if the intent is to document that thread creation can have options

docs/software_requirements/threads.sdoc
```diff
@@ -181,16 +202,309 @@ RELATIONS:
...
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-1-27
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Threads
+TITLE: Querying thread stack usage
+STATEMENT: >>>
+Where thread stack information is enabled, the Zephyr RTOS shall provide a mechanism to determine the amount of stack space a thread has used.
```
parphane, line 402: Couldn't find API related to that. Should this be relocated elsewhere?

docs/software_requirements/thread_scheduling.sdoc
parphane, line 2: Reviewed against: - https://github.com/zephyrproject-rtos/zephyr/blob/main/kernel/sched.c

**Remarks**  

- Performed standard review (not against code/documentation).

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,335 @@ SPDX-License-Identifier: Apache-2.0
...
-UID: ZEP-SRS-2-10
+UID: ZEP-SRS-2-26
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Thread Scheduling
-TITLE: Un-preemptible thread priorities
+TITLE: Preemption status across blocking
 STATEMENT: >>>
-The Zephyr RTOS shall support thread priorities which cannot be preempted by other user threads.
+When a thread that has disabled preemption becomes the current thread again, the Zephyr RTOS shall maintain the thread's non-preemptible status.
```
parphane: Is that requirement necessary? What does "maintain" means in term of logic? If the non-preemptible logic is defined as not preemptible while running, then this requirement is not necessary

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,335 @@ SPDX-License-Identifier: Apache-2.0
...
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-20
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Querying thread preemptibility
+STATEMENT: >>>
+The Zephyr RTOS shall provide an interface to determine whether the currently executing context can be preempted.
```
parphane: Context or thread?

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,335 @@ SPDX-License-Identifier: Apache-2.0
...
+  VALUE: ZEP-SYRS-16
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-21
+STATUS: Draft
+TYPE: Non-Functional
+COMPONENT: Thread Scheduling
+TITLE: Selecting the ready queue implementation
+STATEMENT: >>>
+The Zephyr RTOS shall support selecting the scheduler's ready queue implementation at build time, allowing the selection to trade off memory usage, code size, and scheduling performance as the number of ready threads grows.
```
parphane: Consider leaning to "The Zephyr RTOS shall support selecting the scheduler's ready queue implementation at build time"
And then having requirements define the kinds of scheduler ready queue implementation (standard, EDF...) 
for example 
"The Zephyr RTOS shall support selecting the XXX scheduler ready queue which does YYY"

docs/software_requirements/thread_scheduling.sdoc
```diff
@@ -14,151 +14,335 @@ SPDX-License-Identifier: Apache-2.0
...
+  VALUE: ZEP-SYRS-17
+
+[REQUIREMENT]
+UID: ZEP-SRS-2-22
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Thread Scheduling
+TITLE: Preemption by deferred interrupt processing threads
+STATEMENT: >>>
+Where deferred interrupt processing threads are configured, when a deferred interrupt processing thread becomes ready, the Zephyr RTOS shall preempt the current thread, including a cooperative or preemption-disabled thread or a lower-priority deferred interrupt processing thread, and select it as the current thread.
```
parphane: This requirements (including = not limited to), translates to "Where deferred interrupt processing threads are configured, when a deferred interrupt processing thread becomes ready, the Zephyr RTOS shall preempt the current thread and select it as the current thread."
Which seems incorrect given the lower priority DIP thread mention.

Consider rephrasing to
"Where deferred interrupt processing threads are configured, when a deferred interrupt processing thread becomes ready, the Zephyr RTOS shall preempt any thread that either:
1. Are of a type other than the deferred interrupt processing
2. Are of deferred interrupt processing type and lower-priority"

docs/software_requirements/smp.sdoc
```diff
@@ -0,0 +1,56 @@
...
+  VALUE: ZEP-SYRS-11
+
+[REQUIREMENT]
+UID: ZEP-SRS-34-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Multi Core
+TITLE: Running threads on specific CPUs
+STATEMENT: >>>
+The Zephyr RTOS shall provide an interface for running threads on specific sets of CPUs ( default is 1 CPU).
```
parphane: Default is 1 CPU for what? Which CPU in case of multicores?
Is requirement about number of CPUs or specific CPUs?

Assuming specific CPUs, consider rephrasing:
"The Zephyr RTOS shall provide an interface to specify which CPU core or set of CPU cores thread shall run on"
"When no CPU core or cores are specified for a thread, the Zephyr RTOS shall run that thread on one CPU"

parphane (changes requested):
Reviewed against: 

**Thread:**
1. #24 
2. https://docs.zephyrproject.org/latest/kernel/services/threads/index.html
3. https://docs.zephyrproject.org/latest/doxygen/html/group__thread__apis.html
4. https://github.com/zephyrproject-rtos/zephyr/blob/main/kernel/thread.c

**Scheduling:**
1. #16 
2. https://github.com/zephyrproject-rtos/zephyr/blob/main/kernel/sched.c or 
Performed top down review (no reverse engineering) against requirements properties defined in the requirements checklist

**SMP:**
1. https://docs.zephyrproject.org/latest/doxygen/html/group__arch-smp.html
2. https://github.com/zephyrproject-rtos/zephyr/blob/main/kernel/smp/smp.c
Performed top down review (no reverse engineering)

docs/system_requirements/index.sdoc
nashif: > 1. Can't we merge the 2 requirements into 1 to simplify?
>    It is not clear (starting review), how requirements will choose to link to one or another.

agreed in principle, but SYRS-15 is currently the parent of requirements in several in-flight PRs so retiring the UID inside this PR would break them mid-review. Will look into all of this system vs software and relations as a followup.

> "The Zephyr RTOS shall provide a framework for managing software threads of execution."

adopted.

docs/software_requirements/threads.sdoc
nashif: > * 27, 28, 29, 30, 34 cannot be linked

Those exist in different groups and API docs related to threads. k_thread_stack_space_get (1-27), k_thread_runtime_stats_get/_all_get (1-28/29), stack sentinel/canary config (1-30), k_thread_runtime_stats_enable/disable (1-34)

docs/software_requirements/threads.sdoc
nashif: k_thread_stack_space_get

docs/software_requirements/thread_scheduling.sdoc
nashif: it is, lock persistence across blocking is not derivable from "non-preemptible while running" — a scheduler lock could plausibly drop when the thread blocks; the docs explicitly guarantee it doesn't.

docs/software_requirements/thread_scheduling.sdoc
nashif: "context" is deliberate (k_is_preempt_thread() answers false in ISRs too)

nashif:
@parphane  please take another look, I think I addressed your change requests.

docs/software_requirements/thread_scheduling.sdoc
tobiaskaestner: just a note, no call for action here

docs/software_requirements/thread_scheduling.sdoc
tobiaskaestner: We can go with the informational note inside the User Story for now.

tobiaskaestner (approved):
As draft this looks mergeable to me.

nashif:
@parphane  PTAL

PR #144, opened 2026-06-25
Rework and enhance thread communications

nashif:
- **requirements: convert Thread Communication into an overview document**
- **requirements: add Message Queue system and software requirements**
- **requirements: add Pipe system and software requirements**
- **requirements: add Polling system and software requirements**

### DIFF
--- a/docs/software_requirements/index.sdoc
+++ b/docs/software_requirements/index.sdoc
@@ -56,6 +56,9 @@ FILE: memory_objects.sdoc
 FILE: memory_protection.sdoc
 
 [DOCUMENT_FROM_FILE]
+FILE: message_queue.sdoc
+
+[DOCUMENT_FROM_FILE]
 FILE: mutex.sdoc
 
 [DOCUMENT_FROM_FILE]
--- a/docs/software_requirements/index.sdoc
+++ b/docs/software_requirements/index.sdoc
@@ -62,6 +62,9 @@ FILE: message_queue.sdoc
 FILE: mutex.sdoc
 
 [DOCUMENT_FROM_FILE]
+FILE: pipe.sdoc
+
+[DOCUMENT_FROM_FILE]
 FILE: power_management.sdoc
 
 [DOCUMENT_FROM_FILE]
--- a/docs/software_requirements/index.sdoc
+++ b/docs/software_requirements/index.sdoc
@@ -65,6 +65,9 @@ FILE: mutex.sdoc
 FILE: pipe.sdoc
 
 [DOCUMENT_FROM_FILE]
+FILE: poll.sdoc
+
+[DOCUMENT_FROM_FILE]
 FILE: power_management.sdoc
 
 [DOCUMENT_FROM_FILE]
--- a/docs/software_requirements/message_queue.sdoc
+++ b/docs/software_requirements/message_queue.sdoc
@@ -1,0 +1,297 @@
+[DOCUMENT]
+TITLE: Message Queues
+REQ_PREFIX: ZEP-SRS-31-
+
+[GRAMMAR]
+IMPORT_FROM_FILE: software_requirements.sgra
+
+[TEXT]
+STATEMENT: >>>
+SPDX-License-Identifier: Apache-2.0
+<<<
+
+[[SECTION]]
+TITLE: Message Queue Definition
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Message queue definition at compile time
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to define and initialize a message queue at compile time, for a specified message size, maximum number of messages, and buffer alignment.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Message queue initialization at run time
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to initialize a message queue at run time using a caller-provided buffer, a message size, and a maximum number of messages.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-3
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Message queue initialization at run time with a dynamically allocated buffer
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to initialize a message queue at run time, allocating the message buffer on behalf of the caller for a requested message size and maximum number of messages.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-31
+
+[[/SECTION]]
+
+[[SECTION]]
+TITLE: Sending Messages
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-4
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Sending a message to the back
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism for threads and interrupt service routines to send a message to the back of a message queue.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-5
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Sending a message to the front
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism for threads and interrupt service routines to send a message to the front of a message queue.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-6
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Send timeout
+STATEMENT: >>>
+When sending a message to a message queue, the Zephyr RTOS shall accept a timeout that specifies the maximum time to wait for space to become available.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-7
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Send timeout expiry
+STATEMENT: >>>
+When sending a message to a message queue, the Zephyr RTOS shall return an error indicating that the message was not sent when either of the following conditions is satisfied:
+
+- the caller is a thread and space does not become available in the message queue within the specified maximum time
+- the caller is an interrupt service routine and no space is available in the message queue
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-31
+
+[[/SECTION]]
+
+[[SECTION]]
+TITLE: Receiving Messages
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-8
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Receiving a message
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism for threads and interrupt service routines to receive and remove the message at the front of a message queue.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-9
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Message retrieval order
+STATEMENT: >>>
+The Zephyr RTOS shall return messages from a message queue to receivers in front-to-back order: messages sent to the back are received in the order in which they were sent, and a message sent to the front is received before the messages already in the queue.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-10
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Receive timeout
+STATEMENT: >>>
+When receiving a message from a message queue, the Zephyr RTOS shall accept a timeout that specifies the maximum time to wait for a message to become available.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-11
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Receive timeout expiry
+STATEMENT: >>>
+When receiving a message from a message queue, the Zephyr RTOS shall return an error indicating that no message was received when either of the following conditions is satisfied:
+
+- the caller is a thread and no message becomes available in the message queue within the specified maximum time
+- the caller is an interrupt service routine and no message is available in the message queue
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-31
+
+[[/SECTION]]
+
+[[SECTION]]
+TITLE: Inspecting and Managing a Message Queue
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-12
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Peeking at a message
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism for threads and interrupt service routines to read the message at the front of a message queue without removing it.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-13
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Peeking at a message by index
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism for threads and interrupt service routines to read a message at a specified index of a message queue without removing it.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-14
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Purging a message queue
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to discard all messages currently in a message queue.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-19
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Purging with blocked senders
+STATEMENT: >>>
+When a message queue is purged, the Zephyr RTOS shall unblock any threads waiting to send a message to that queue and return an error to them indicating that the message was not sent.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-15
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Querying used message count
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to query the number of messages currently in a message queue.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-16
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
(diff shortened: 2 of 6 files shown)


docs/software_requirements/message_queue.sdoc
```diff
@@ -0,0 +1,252 @@
...
+TITLE: Message Queue Definition
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Message queue definition at compile time
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to define and initialize a message queue at compile time.
```
anobli: Should we not also talk about arguments as this is done in the next requirements ?

docs/software_requirements/message_queue.sdoc
```diff
@@ -0,0 +1,252 @@
...
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-3
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Message queue buffer dynamic allocation
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to initialize a message queue with a buffer that is dynamically allocated for a requested message size and maximum number of messages.
```
anobli: I found the sentence confusing. When I read the sentence, I am not sure if this the function that does the allocation of buffer, or, if the buffer is given as an argument.
The first option is the more obvious and is confirmed by the title but I think we should update the sentence to make it more obvious.

docs/software_requirements/message_queue.sdoc
```diff
@@ -0,0 +1,252 @@
...
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-3
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Message queue buffer dynamic allocation
```
anobli: This let's think that we only allocate a buffer whereas we also initialize the message queue

docs/software_requirements/message_queue.sdoc
```diff
@@ -0,0 +1,252 @@
...
+TITLE: Sending Messages
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-4
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Sending a message to the back
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism for threads and interrupt service routines to send a message to the end of a message queue.
```
anobli: nit: consistency with the title: maybe we should replace end by back

docs/software_requirements/message_queue.sdoc
```diff
@@ -0,0 +1,252 @@
...
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-6
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Send timeout
+STATEMENT: >>>
+When sending a message to a message queue, the Zephyr RTOS shall accept a timeout that specifies the maximum time to wait for space to become available.
```
anobli, line 93: The timeout can only be set when we are sending a message from a thread, not an ISR. It should mentioned.

docs/software_requirements/message_queue.sdoc
```diff
@@ -0,0 +1,252 @@
...
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-9
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Message retrieval order
+STATEMENT: >>>
+When a message is received from a message queue, the Zephyr RTOS shall return messages in first-in, first-out order, except for messages that were sent to the front of the queue.
```
anobli: I find the exception a bit confusing. It would be more obvious without it in my opinion even if I understand the purpose. Maybe just saying that we  always take the first in the front of queue

docs/software_requirements/message_queue.sdoc
```diff
@@ -0,0 +1,252 @@
...
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-10
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Receive timeout
+STATEMENT: >>>
+When receiving a message from a message queue, the Zephyr RTOS shall accept a timeout that specifies the maximum time to wait for a message to become available.
```
anobli, line 153: Same as for sensing, I think timeout is only applicable for threads, not ISR

docs/software_requirements/message_queue.sdoc
```diff
@@ -0,0 +1,252 @@
...
+TITLE: Inspecting and Managing a Message Queue
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-12
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Peeking at a message
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to read the message at the front of a message queue without removing it.
```
anobli: I think we should mention that it could be done from ISR or thread

docs/software_requirements/message_queue.sdoc
```diff
@@ -0,0 +1,252 @@
...
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-13
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Peeking at a message by index
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to read a message at a specified index of a message queue without removing it.
```
anobli: Same here.

docs/software_requirements/message_queue.sdoc
```diff
@@ -0,0 +1,252 @@
...
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-14
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Purging a message queue
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to discard all messages currently in a message queue.
```
anobli, line 213: maybe we should also mention that unblock any thread waiting for a message (maybe as part of another rquirements)

docs/software_requirements/message_queue.sdoc
```diff
@@ -0,0 +1,252 @@
...
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-17
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Releasing a dynamically allocated message queue buffer
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to release the buffer of a message queue that was dynamically allocated.
```
anobli, line 278: May be we should say that is only for an empty message queue

docs/software_requirements/index.sdoc
```diff
@@ -55,6 +55,15 @@ FILE: power_management.sdoc
 [DOCUMENT_FROM_FILE]
 FILE: thread_communication.sdoc
```
parphane: Agree with leaving the file to have 1 generic requirement that mentions the type of communications we want.
In the future, we may consider a "generic" RTOS requirements file to compile such requirements, instead of having 1 file with 1 requirement.
However, consider removing the file altogether, as it adds maintenance effort and contains no requirement, just text with inference to mechanisms, which may or may not be ambiguously seen as a requirement

docs/software_requirements/message_queue.sdoc
parphane: Closing, fixed by having end and end in both title and req.
However, consider consistency with Queue where terminology was discussed and agreed to use front and back (https://github.com/zephyrproject-rtos/reqmgmt/blob/main/docs/software_requirements/queues.sdoc)

docs/software_requirements/message_queue.sdoc
```diff
@@ -0,0 +1,291 @@
...
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-7
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Send timeout expiry
+STATEMENT: >>>
+When space does not become available in a message queue within the specified time, the Zephyr RTOS shall return an error indicating that the message was not sent.
```
parphane: Consider reusing verbiage from above requirement, as in, replace "specified time" by "specified maximum time" or something similar. Also consider adding "When sending a message to a message queue" to provide proper context.
Could consider using a bullet point structure for ISRs:
When sending a message to a message queue, the Zephyr RTOS shall return an error indicating that the message was not sent when any of the following conditions is satisfied:
- the caller is a thread and space does not become available in the message queue within the specified time
- the caller is an ISR and no space is available in the message queue

docs/software_requirements/message_queue.sdoc
parphane: The timeout is "accepted" in both cases, so that requirement is fine for ISRs/Threads.
But see proposition for req 7 for the logic

docs/software_requirements/message_queue.sdoc
```diff
@@ -0,0 +1,291 @@
...
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-9
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Message retrieval order
+STATEMENT: >>>
+When messages are sent to the end of a message queue, the Zephyr RTOS shall return them to receivers in the order in which they were sent (first-in, first-out).
```
parphane: Consider removing req 9 and 18 altogether.
The mechanism is simple here, there is a function to receive the message at the front of the message queue.
Then there are mechanisms to put in front or back.

docs/software_requirements/message_queue.sdoc
parphane: The timeout is "accepted" in both cases, so that requirement is fine for ISRs/Threads.
See below proposition for req 11 for the logic.

docs/software_requirements/message_queue.sdoc
```diff
@@ -0,0 +1,291 @@
...
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-11
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Receive timeout expiry
+STATEMENT: >>>
+When no message becomes available in a message queue within the specified time, the Zephyr RTOS shall return an error indicating that no message was received.
```
parphane: Similar comment to above, consider rephrasing.
Could consider using a bullet point structure for ISRs:
When receiving a message from a message queue, the Zephyr RTOS shall return an error indicating that the message was not sent when any of the following conditions is satisfied:
- the caller is a thread and no message becomes available in the message queue within the specified time
- the caller is an ISR and no message is available in the message queue

docs/software_requirements/message_queue.sdoc
parphane: Fixed by ZEP-SRS-31-19

docs/software_requirements/message_queue.sdoc
```diff
@@ -0,0 +1,291 @@
...
+  VALUE: ZEP-SYRS-31
+
+[REQUIREMENT]
+UID: ZEP-SRS-31-20
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Message Queues
+TITLE: Releasing a buffer with waiting threads
+STATEMENT: >>>
+If threads are waiting on a message queue, then the Zephyr RTOS shall not release the message queue's buffer and shall return an error indicating that the message queue is in use.
```
parphane: Need to clarify the "When", else this is applicable anytime.
Proposition:
"When releasing the buffer of a message queue, if threads are waiting on the message queue, the Zephyr RTOS shall not release the message queue's buffer and shall return an error indicating that the message queue is in use.

docs/software_requirements/message_queue.sdoc
```diff
@@ -0,0 +1,291 @@
+[DOCUMENT]
```
parphane, line 1: Design constraints
6, 7
Macros
1 - #define 	K_MSGQ_DEFINE(q_name, q_msg_size, q_max_msgs, q_align)

Functions
2 - void 	k_msgq_init (struct k_msgq *msgq, char *buffer, size_t msg_size, uint32_t max_msgs)
3 - int 	k_msgq_alloc_init (struct k_msgq *msgq, size_t msg_size, uint32_t max_msgs)
17, 20 - int 	k_msgq_cleanup (struct k_msgq *msgq)
4 - int 	k_msgq_put (struct k_msgq *msgq, const void *data, k_timeout_t timeout)
5 - int 	k_msgq_put_front (struct k_msgq *msgq, const void *data)
8, 9, 18, 11 - int 	k_msgq_get (struct k_msgq *msgq, void *data, k_timeout_t timeout)
12 - int 	k_msgq_peek (struct k_msgq *msgq, void *data)
13 - int 	k_msgq_peek_at (struct k_msgq *msgq, void *data, uint32_t idx)
14, 19 - void 	k_msgq_purge (struct k_msgq *msgq)
16 - uint32_t 	k_msgq_num_free_get (struct k_msgq *msgq)
void 	k_msgq_get_attrs (struct k_msgq *msgq, struct k_msgq_attrs *attrs)
15 - uint32_t 	k_msgq_num_used_get (struct k_msgq *msgq)

Missing requirement for "k_msgq_get_attrs"

docs/software_requirements/pipe.sdoc
```diff
@@ -0,0 +1,141 @@
...
+  VALUE: ZEP-SYRS-32
+
+[REQUIREMENT]
+UID: ZEP-SRS-32-9
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Pipes
+TITLE: Operations on a closed pipe
+STATEMENT: >>>
+If a read or write operation is attempted on a closed pipe, then the Zephyr RTOS shall return an error indicating that the pipe is closed.
```
parphane: Consider replacing "If", by "When"

docs/software_requirements/message_queue.sdoc
parphane: Fixed by addition of req 20

docs/software_requirements/pipe.sdoc
```diff
@@ -0,0 +1,141 @@
...
+  VALUE: ZEP-SYRS-32
+
+[REQUIREMENT]
+UID: ZEP-SRS-32-5
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Pipes
+TITLE: Pipe write timeout
+STATEMENT: >>>
+When writing to a pipe, the Zephyr RTOS shall accept a timeout that specifies the maximum time to wait for space to become available.
```
parphane, line 72: 5 and 6, for consistency with message queue, are missing a requirement for the timer expiry logic.

docs/software_requirements/pipe.sdoc
```diff
@@ -0,0 +1,141 @@
+[DOCUMENT]
```
parphane, line 1: Macros
1 - #define 	K_PIPE_DEFINE(name, pipe_buffer_size, pipe_align)

Functions
2 - void 	k_pipe_init (struct k_pipe *pipe, uint8_t *buffer, size_t buffer_size)
3, 5, 9, 10 - int 	k_pipe_write (struct k_pipe *pipe, const uint8_t *data, size_t len, k_timeout_t timeout)
4, 6, 9, 10 - int 	k_pipe_read (struct k_pipe *pipe, uint8_t *data, size_t len, k_timeout_t timeout)
7 - void 	k_pipe_reset (struct k_pipe *pipe)
8 - void 	k_pipe_close (struct k_pipe *pipe)

Requirements are covered by APIs and vice versa.

docs/software_requirements/poll.sdoc
```diff
@@ -0,0 +1,151 @@
...
+  VALUE: ZEP-SYRS-33
+
+[REQUIREMENT]
+UID: ZEP-SRS-33-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Polling
+TITLE: Supported poll event conditions
+STATEMENT: >>>
+The Zephyr RTOS shall support poll events that wait for the availability of a semaphore, a FIFO, a message queue, a pipe, and a poll signal.
```
parphane, line 36: If this is for a MAcro/Function please specify "shall provide a mechanism/compile-time mechanism"

docs/software_requirements/poll.sdoc
```diff
@@ -0,0 +1,151 @@
...
+TITLE: Poll Signals
+
+[REQUIREMENT]
+UID: ZEP-SRS-33-5
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Polling
+TITLE: Poll signal object
+STATEMENT: >>>
+The Zephyr RTOS shall provide a poll signal object that a thread can wait on as a poll event.
```
parphane: Are we providing 1 object or a mechanism to initialize objects that threads can wait on as they poll?

docs/software_requirements/poll.sdoc
```diff
@@ -0,0 +1,151 @@
...
+  VALUE: ZEP-SYRS-33
+
+[REQUIREMENT]
+UID: ZEP-SRS-33-8
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Polling
+TITLE: Persistence of a raised poll signal
+STATEMENT: >>>
+While a poll signal has been raised and not reset, the Zephyr RTOS shall keep the signal in the signaled state.
```
parphane: Consider removing, there are requirements dictating signaled state change already (6 for reset, 7 to decide what value to raise).
If the objective is to prevent a signal to be "reraised", then consider modifying req 7 logic to prevent changing the value if it is in any other state than "unsignaled"

docs/software_requirements/poll.sdoc
```diff
@@ -0,0 +1,151 @@
+[DOCUMENT]
```
parphane, line 1: Macros
5 - #define 	K_POLL_SIGNAL_INITIALIZER(obj)
#define 	K_POLL_EVENT_INITIALIZER(_event_type, _event_mode, _event_obj)
#define 	K_POLL_EVENT_STATIC_INITIALIZER(_event_type, _event_mode, _event_obj, event_tag)

Functions
void 	k_poll_event_init (struct k_poll_event *event, uint32_t type, int mode, void *obj)
1?, 2?, 3, 4 - int 	k_poll (struct k_poll_event *events, int num_events, k_timeout_t timeout)
6 - void 	k_poll_signal_init (struct k_poll_signal *sig)
9 - void 	k_poll_signal_reset (struct k_poll_signal *sig)
10 - void 	k_poll_signal_check (struct k_poll_signal *sig, unsigned int *signaled, int *result)
7, 8 - int 	k_poll_signal_raise (struct k_poll_signal *sig, int result)

It is unclear to me what requirements 1 2 are linked to.
Once clarified, let's check what requirements are covered by K_POLL_EVENT_INITIALIZER, K_POLL_EVENT_STATIC_INITIALIZER, k_poll_event_init

parphane (changes requested):
Performed review on changes:
1. thread_communication update to list communication mechanisms and have 1 file per API. But I would consider removing the thread_comm file altogether
2. system requirements update to add mqueue, pipe and poll APIs files
3. software requirements reviewed individually and comments left

docs/software_requirements/poll.sdoc
nashif: deliberately not API-shaped, it constrains which condition types the wait mechanism (33-1) supports.

docs/software_requirements/poll.sdoc
nashif: Reworded rather than removed, because the property it carries isn't derivable from 33-6/33-7/33-9: a raised signal is level-triggered: a poll wait that begins after the raise is still reported ready, until the signal is explicitly reset.

docs/software_requirements/poll.sdoc
nashif: 33-2 → the supported event condition types (a constraint on 33-1, not an API)

33-1/33-3/33-4 → k_poll

docs/software_requirements/index.sdoc
nashif: Kept for now, per the agreement in the other thread: it anchors the Thread Communication section and inventories which mechanisms have their own documents, and it contains no requirement that could be ambiguously interpreted (it's a [TEXT] block only).

docs/software_requirements/poll.sdoc
parphane: Fine for now, since this is a top down exercise, any missing traceability to K_POLL_EVENT_INITIALIZER, K_POLL_EVENT_STATIC_INITIALIZER or k_poll_event_init can be resolved in future analysis for coverage gaps.

docs/software_requirements/poll.sdoc
parphane: Assuming the intent is to tell that poll signal object can be waited on.

docs/software_requirements/message_queue.sdoc
parphane: Not removed as part of rework (not an issue). Closing remark

docs/software_requirements/message_queue.sdoc
parphane: Fixed

docs/software_requirements/message_queue.sdoc
parphane: As mentioned below, will be fixed during coverage analysis as needed. Review is performed on requirements, not against the code/APIs

docs/software_requirements/pipe.sdoc
parphane: Fixed in 11 and 12

parphane (approved):
Reviewed rework 
Also reviewed rework from anobli
All comments closed 
Approving for merge

PR #145, opened 2026-06-25
Fix timer / condvar requirements

nashif:
- **requirements: correct Timers requirement types and wording**
- **requirements: fix Condition Variables parent links and sync wording**


parphane (approved):
Reviewed:
1. Trace to System requirement
2. Changes in requirements (no functional changes, only typo/clarification)

PR #146, opened 2026-06-25
Mutex requirements

nashif:
- **requirements: correct Mutex umbrella requirement**
- **requirements: add missing Mutex software requirements**


parphane (approved):
Performed review:

1. Agreed with change on "Mutex Kernel Object", this SW req is about Mutex. If we want a generic requirement, we should make one at system level.
2. Properly traced to sys req ZEP-SYRS-13
3. 4 functions covered by requirements, internal function logic defined without ambiguity
4. No checklist violation https://docs.zephyrproject.org/latest/safety/safety_requirements/safety_requirements_checklist.html

PR #148, opened 2026-06-25
requirements: build out System Initialization requirements

nashif:
The System Initialization document was a TBD stub with no requirements.
Add a System requirement (ZEP-SYRS-34) for system initialization and
replace the stub with 9 software requirements (ZEP-SRS-12-*) covering
registration of initialization functions, the initialization levels,
initialization ordering by level and by priority, execution relative to
kernel-service availability, device initialization, and starting the
application after initialization completes.

The requirements reflect the currently implemented SYS_INIT and boot
sequence behavior following the Route 3s approach and link to the
ZEP-SYRS-34 system parent.

### DIFF
--- a/docs/software_requirements/system_initialization.sdoc
+++ b/docs/software_requirements/system_initialization.sdoc
@@ -10,7 +10,184 @@ STATEMENT: >>>
 SPDX-License-Identifier: Apache-2.0
 <<<
 
-[TEXT]
+[REQUIREMENT]
+UID: ZEP-SRS-12-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Registering initialization functions
 STATEMENT: >>>
-TBD
+The Zephyr RTOS shall provide a mechanism to register a function to be invoked automatically during system initialization.
 <<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-11
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Named registration of initialization functions
+STATEMENT: >>>
+The Zephyr RTOS shall support associating a distinct name with each registration of an initialization function, allowing the same function to be registered for initialization more than once.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Initialization levels
+STATEMENT: >>>
+When registering a function for initialization, an initialization level shall be provided, and the Zephyr RTOS shall associate the function with that level.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-12
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Initialization priority
+STATEMENT: >>>
+When registering a function for initialization, a priority within the initialization level shall be provided, and the Zephyr RTOS shall associate the function with that priority.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-13
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Defined initialization levels
+STATEMENT: >>>
+The Zephyr RTOS shall organize system initialization into an ordered sequence of initialization levels: early hardware initialization, pre-kernel initialization performed before kernel services are available, post-kernel initialization performed with kernel services available, and application initialization performed immediately before the application entry point is invoked.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-14
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Querying an initialization level ordinal
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to obtain the ordinal number of an initialization level.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-3
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Initialization order by level
+STATEMENT: >>>
+During system initialization, the Zephyr RTOS shall invoke registered initialization functions level by level, in the defined order of initialization levels.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-4
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Initialization order by priority
+STATEMENT: >>>
+Within an initialization level, the Zephyr RTOS shall invoke registered initialization functions in ascending order of their registered priority.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-5
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Initialization before kernel services
+STATEMENT: >>>
+The Zephyr RTOS shall execute the early hardware and pre-kernel initialization levels in a single-threaded context before the kernel scheduler is started.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-6
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Initialization after kernel services
+STATEMENT: >>>
+The Zephyr RTOS shall execute the post-kernel and application initialization levels in the context of a kernel thread after the kernel scheduler has started, with kernel services available.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-7
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Device initialization
+STATEMENT: >>>
+The Zephyr RTOS shall initialize devices as part of system initialization, invoking each device's initialization function according to its assigned initialization level and priority.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-8
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Application start after initialization
+STATEMENT: >>>
+The Zephyr RTOS shall complete all initialization levels before invoking the application entry point.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-9
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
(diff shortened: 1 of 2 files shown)


docs/software_requirements/system_initialization.sdoc
```diff
@@ -10,7 +10,106 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-7
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Device initialization
+STATEMENT: >>>
+The Zephyr RTOS shall initialize configured device drivers during system initialization according to their assigned initialization level and priority.
```
parphane: Is this requirement necessary? Aren't device driver init functions treated like any other functions registered for init? (I did not check code/implementation)

docs/software_requirements/system_initialization.sdoc
```diff
@@ -10,7 +10,106 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Initialization levels
+STATEMENT: >>>
+The Zephyr RTOS shall allow each registered initialization function to be assigned to one of the defined initialization levels.
```
parphane: Is level optional? Consider replacing allow by 
"When registering functions for initialization, an initialization level shall be provided, and the Zephyr RTOS shall associate it to the function"
This can then cover both INIT and INIT_NAMED
Similarly for the missing req on priority
"When registering functions for initialization, a priority level shall be provided, and the Zephyr RTOS shall associate it to the function"

docs/software_requirements/system_initialization.sdoc
parphane, line 1: Design constraints
2, 3, 4, 5, 6

Macros
- #define INIT_LEVEL_ORD(level)
1 - #define SYS_INIT(init_fn, level, prio)
 - #define SYS_INIT_NAMED(name, init_fn_, level, prio)

Missing requirements for 2 APIs SYS_INIT_NAMED and INIT_LEVEL_ORD
- "The Zephyr RTOS shall provide a mechanism to register a function, with an associated name, to be invoked automatically during system initialization."
- "The Zephyr RTOS shall provide a mechanism to obtain an initialization level's ordinal number."
Missing requirement, like ZEP-SRS-12-2, to allow priority to be assigned when registering
- "When registering functions for initialization, a priority level shall be provided, and the Zephyr RTOS shall associate it to the function"
Missing requirement, like ZEP-SRS-12-2, to allow priority to be assigned when registering

docs/software_requirements/system_initialization.sdoc
```diff
@@ -10,7 +10,106 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-3
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Initialization order by level
+STATEMENT: >>>
+During system initialization, the Zephyr RTOS shall invoke registered initialization functions in ascending order of their initialization level.
```
parphane: Ambiguous since no initialization levels are defined nor a number assigned to them.
Also maybe redundant with requirements 5 and 6 that re-define initialization order/constraints.

docs/software_requirements/system_initialization.sdoc
parphane: No initialization levels are formally defined but some are referenced.

EARLY - No requirement attached
PRE_KERNEL_1, PRE_KERNEL_2 - Req. 5 attached 
POST_KERNEL, APPLICATION, SMP - Req. 6 attached

docs/software_requirements/system_initialization.sdoc
```diff
@@ -10,7 +10,106 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-5
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Initialization before kernel services
+STATEMENT: >>>
+The Zephyr RTOS shall execute the PRE_KERNEL initialization levels before kernel services are available.
```
parphane: Is this requirement needed? We are already defining an order in 3.
If we are waiting on an outcome, should it be rephrased as "The Zephyr RTOS shall execute the POST_KERNEL" when and only when the PRE_KERNEL_1 and PRE_KERNEL_2 initialization level functions <HAVE DONE/RETUNED SOMETHING>?

Where is the list of Kernel services / requirements defined? 
In absence of Kernel services requirements kernel services list and meaning of "available" are ambiguous.

parphane (changes requested):
Reviewed:
1. System requirements
2. Software requirements

OS Services / Init: https://docs.zephyrproject.org/latest/doxygen/html/group__sys__init.html
Code: https://github.com/zephyrproject-rtos/zephyr/blob/f45025108cb0848e0d7419159a3ed76d9fcc0459/include/zephyr/init.h

nashif:
@parphane  please take another look

@zephyrproject-rtos/safety  FYI

docs/software_requirements/system_initialization.sdoc
```diff
@@ -10,7 +10,184 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-11
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Named registration of initialization functions
+STATEMENT: >>>
+The Zephyr RTOS shall support associating a distinct name with each registration of an initialization function, allowing the same function to be registered for initialization more than once.
```
anobli, line 33: I am not sure how to deal with that one.
Currently this doesn't make explicit that we have an macro for that and remains ambiguous.
I guess we should probably define the mecanism we want.

docs/software_requirements/system_initialization.sdoc
```diff
@@ -10,7 +10,184 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-13
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Defined initialization levels
+STATEMENT: >>>
+The Zephyr RTOS shall organize system initialization into an ordered sequence of initialization levels: early hardware initialization, pre-kernel initialization performed before kernel services are available, post-kernel initialization performed with kernel services available, and application initialization performed immediately before the application entry point is invoked.
```
anobli: I am wondering if we should not break this in multiple requirements. Especially, if we consider SMP level (which is missing) that is optional and would require a "where" keyword to be described.

docs/software_requirements/system_initialization.sdoc
```diff
@@ -10,7 +10,184 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-3
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Initialization order by level
+STATEMENT: >>>
+During system initialization, the Zephyr RTOS shall invoke registered initialization functions level by level, in the defined order of initialization levels.
```
anobli: I am sure that `during` is valid regarding EARS.

docs/software_requirements/system_initialization.sdoc
```diff
@@ -10,7 +10,184 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-5
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Initialization before kernel services
+STATEMENT: >>>
+The Zephyr RTOS shall execute the early hardware and pre-kernel initialization levels in a single-threaded context before the kernel scheduler is started.
```
anobli: Should we not also mention that we the ISR stack ?

docs/software_requirements/system_initialization.sdoc
```diff
@@ -10,7 +10,184 @@ STATEMENT: >>>
...
 
-[TEXT]
+[REQUIREMENT]
+UID: ZEP-SRS-12-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Registering initialization functions
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to register a function to be invoked automatically during system initialization.
```
anobli: Should not also define what is an initialization function ?
I mean, does it take arguments, does it return something ?

docs/software_requirements/system_initialization.sdoc
```diff
@@ -10,7 +10,184 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-4
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Initialization order by priority
+STATEMENT: >>>
+Within an initialization level, the Zephyr RTOS shall invoke registered initialization functions in ascending order of their registered priority.
```
anobli, line 111: What does happen in case of a function fails ?

docs/software_requirements/system_initialization.sdoc
```diff
@@ -10,7 +10,184 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-7
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Device initialization
+STATEMENT: >>>
+The Zephyr RTOS shall initialize devices as part of system initialization, invoking each device's initialization function according to its assigned initialization level and priority.
```
tobiaskaestner, line 150: What about deferred device driver initialization? The way the requirement is stated, it is total and would not allow for deferred initialization.

tobiaskaestner (commented):
Only one comment from my side to double check, regarding deferred init of device drivers.

docs/software_requirements/system_initialization.sdoc
anobli: or maybe to remain consistent with priority and level:
`When registering a function for initialization, a distinct name shall be provided ...`

docs/software_requirements/system_initialization.sdoc
```diff
@@ -10,7 +10,184 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-12
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: Initialization priority
+STATEMENT: >>>
+When registering a function for initialization, a priority within the initialization level shall be provided, and the Zephyr RTOS shall associate the function with that priority.
```
anobli, line 59: May be this is implementation and thus could be ignored: should we not define the priority range (0 to 999)?

docs/software_requirements/system_initialization.sdoc
anobli, line 72: Here, I would just name the init levels and say that this respects the same order. Information such a level is performed with kernel services available should be done in a separate requirement, as this already done by  ZEP-SRS-12-5 and  ZEP-SRS-12-6.

anobli: Also, what about pre_kernel_1 and pre_kernel_2? Although they are same (I mean, run in the same context, etc), pre_kernel_1 still run before pre_kernel_2 if understand correctly.

docs/software_requirements/system_initialization.sdoc
anobli: also, what does happen if we have to function registered with the same priority ?

docs/software_requirements/system_initialization.sdoc
anobli, line 124: I think  it worth mentioning that for that init level, IRQ is disabled and init function will never be interrupted.

docs/software_requirements/system_initialization.sdoc
```diff
@@ -10,7 +10,184 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-12-9
+STATUS: Draft
+TYPE: Functional
+COMPONENT: System Initialization
+TITLE: SMP initialization
+STATEMENT: >>>
+Where symmetric multiprocessing is enabled, the Zephyr RTOS shall bring up the secondary CPUs during system initialization, before the application is started.
```
anobli, line 176: Here `before the application is started` is confusing.
Despite 12-8 make it obvious, application might be read as application level instead of application entry point. And this doesn't clearly explain that this is run before application entry point but after the application level init.

PR #150, opened 2026-06-25
semaphore requirement additions / enhancements

nashif:
- **requirements: correct Semaphore reset error semantics**
- **requirements: add Semaphore give-at-max and ISR usage requirements**

### DIFF
--- a/docs/software_requirements/semaphore.sdoc
+++ b/docs/software_requirements/semaphore.sdoc
@@ -247,3 +247,29 @@ then the Zephyr RTOS shall return an error indicating invalid values.
 RELATIONS:
 - TYPE: Parent
   VALUE: ZEP-SYRS-14
+
+[REQUIREMENT]
+UID: ZEP-SRS-5-19
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Semaphore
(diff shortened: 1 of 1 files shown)


parphane (approved):
Reviewed changes against: 
- https://docs.zephyrproject.org/latest/doxygen/html/group__semaphore__apis.html
- https://github.com/zephyrproject-rtos/zephyr/blob/main/kernel/sem.c

1. ZEP-SRS-5-17: Non functional change accepted
2. ZEP-SRS-5-19: Ensuring counter does not exceed maximum, in conjunction with ZEP-SRS-5-13
3. ZEP-SRS-5-20: Accepting as is for now but could be expanded like is done in queues to describe the timeout logic thread vs ISR

Also performed unrelated existing requirements review:
- ZEP-SRS-5-14: "When there are threads waiting on the semaphore, the highest-priority waiting thread shall be unblocked and acquire the semaphore." could be interpreted as any time there are threads waiting, then we should unblock one. Also conflicts in logic with ZEP-SRS-5-8
- ZEP-SRS-5-18: "invalid" is ambiguous
Will open another issue to fix

parphane:
Created #157 to log unrelated review remarks& suggestions

Issue #138, opened 2026-06-15
Queue parent relations point to the Condition Variables parent


RICCIARDI-Adrien:
The Queue parent relations `ZEP-SYRS-21` seem to belong to the `Condition Variables`, which I think is not the intended target.

RICCIARDI-Adrien:
Fixed by 0d068c0.

PR #153, opened 2026-07-07
queues-requirements: Fix the parent relation values

RICCIARDI-Adrien:
The parent relation values were pointing to the top requirement about condition variables instead of the one for the queues.

This addresses the issue https://github.com/zephyrproject-rtos/reqmgmt/issues/138.

### DIFF
--- a/docs/software_requirements/fifos.sdoc
+++ b/docs/software_requirements/fifos.sdoc
@@ -134,16 +134,3 @@ The Zephyr RTOS shall provide a mechanism to peek at the data item at the back o
 RELATIONS:
 - TYPE: Parent
   VALUE: ZEP-SYRS-24
-
-[REQUIREMENT]
-UID: ZEP-SRS-24-11
-STATUS: Draft
-TYPE: Functional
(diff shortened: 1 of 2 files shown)


docs/software_requirements/fifos.sdoc
nashif: this whole requirement needs to move to queues. no reason to keep it under fifos.

RICCIARDI-Adrien: Indeed, seems to be a left-over. This requirement is already existing for FIFOs (see https://github.com/zephyrproject-rtos/reqmgmt/blob/a5c154d946aa1cb4efee39e46fc7a7c725a02f86/docs/software_requirements/fifos.sdoc#L9) and for queues (see https://github.com/zephyrproject-rtos/reqmgmt/blob/a5c154d946aa1cb4efee39e46fc7a7c725a02f86/docs/software_requirements/queues.sdoc#L9), so I just removed it.

nashif (approved):

PR #156, opened 2026-07-23
atomic service enhancements / fixes / additions

nashif:
- **requirements: atomic: fix wording, titles and grammar**
- **requirements: atomic: simplify parent relations**
- **requirements: atomic: add usability from threads and ISRs**


parphane (approved):
Reviewed non-functional changes
1. Clarification / Abstraction of concepts (HW vs Concept or How vs What)
2. Trace updates to SYSRS instead of design constraints (SW req 1, 2, 3, 4, 5, 6) 
3. Clarification of performing operation against a provided value

Noting that ZEP-SRS-26-40 is somewhat redundant but consistent with ZEP-SRS-26-6

PR #158, opened 2026-07-25
requirements: add Events system and software requirements

nashif:
Add a System requirement (ZEP-SYRS-27) describing the Zephyr event
object synchronization mechanism, and a new Events software requirements
document (ZEP-SRS-27-*) refining it into 15 atomic, implementation-free
requirements covering event object definition, delivering events
(post/set/set-masked/clear), and waiting for events (any/all, timeout,
reset, reported results).

The requirements reflect the currently implemented k_event API behavior
following the Route 3s approach. All software requirements link to the
ZEP-SYRS-27 system parent, and the new document is registered in the
software requirements index. UIDs were generated with
"strictdoc manage auto-uid".

### DIFF
--- a/docs/software_requirements/events.sdoc
+++ b/docs/software_requirements/events.sdoc
@@ -1,0 +1,221 @@
+[DOCUMENT]
+TITLE: Events
+REQ_PREFIX: ZEP-SRS-27-
+
+[GRAMMAR]
+IMPORT_FROM_FILE: software_requirements.sgra
+
+[TEXT]
+STATEMENT: >>>
+SPDX-License-Identifier: Apache-2.0
+<<<
+
+[[SECTION]]
+TITLE: Event Object Definition
+
+[REQUIREMENT]
+UID: ZEP-SRS-27-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Events
+TITLE: Event object definition at compile time
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to define and initialize an event object at compile time.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-27
+
+[REQUIREMENT]
+UID: ZEP-SRS-27-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Events
+TITLE: Event object definition at run time
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to initialize an event object at run time.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-27
+
+[REQUIREMENT]
+UID: ZEP-SRS-27-3
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Events
+TITLE: Number of distinct events
+STATEMENT: >>>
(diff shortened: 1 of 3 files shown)


docs/software_requirements/events.sdoc
```diff
@@ -0,0 +1,221 @@
...
+  VALUE: ZEP-SYRS-27
+
+[REQUIREMENT]
+UID: ZEP-SRS-27-3
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Events
+TITLE: Number of distinct events
+STATEMENT: >>>
+An event object shall track up to 32 distinct events.
```
anobli: Inconsistency! I would suggest  `The Zephyr RTOS shall provide an event object capable of tracking up to 32 distinct events.`

docs/software_requirements/events.sdoc
```diff
@@ -0,0 +1,221 @@
...
+  VALUE: ZEP-SYRS-27
+
+[REQUIREMENT]
+UID: ZEP-SRS-27-6
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Events
+TITLE: Setting and clearing events using a mask
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to set and clear individual events tracked by an event object as determined by a mask.
```
anobli: I think this one need to be more detailed. I think we need to explain that only bit set in mask are getting updated.

docs/software_requirements/events.sdoc
```diff
@@ -0,0 +1,221 @@
...
+  VALUE: ZEP-SYRS-27
+
+[REQUIREMENT]
+UID: ZEP-SRS-27-8
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Events
+TITLE: Reporting the previous events
+STATEMENT: >>>
+When events are posted to or set in an event object, the Zephyr RTOS shall report the value of the events that were tracked by the event object before the operation.
```
anobli: Hum, how this is written, it doesn't seem to apply to ZEP-SRS-27-6 and ZEP-SRS-27-7.
I think we could says something like: `When events tracked by an event object are modified ...`

docs/software_requirements/events.sdoc
```diff
@@ -0,0 +1,221 @@
...
+  VALUE: ZEP-SYRS-27
+
+[REQUIREMENT]
+UID: ZEP-SRS-27-9
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Events
+TITLE: Unpending threads whose conditions are met
+STATEMENT: >>>
+When events posted to or set in an event object satisfy the waiting condition of a thread waiting on that event object, the Zephyr RTOS shall unpend that thread.
```
anobli: Same here.

docs/software_requirements/events.sdoc
tobiaskaestner: Fine with it as is for me. The semantics of the mask is implementation detail to me. Although natural expectation would be to select affected events by bits set to one in the mask.

tobiaskaestner (approved):
Comments are valuable and worth addressing, yet I don't consider them blocking right now.

nashif:
> Comments are valuable and worth addressing, yet I don't consider them blocking right now.

addressed

PR #161, opened 2026-07-27
Kernel Timing

nashif:
- **requirements: add Kernel Timing system and software requirements**
- **requirements: timers: correct wording and add cleanup and observer coverage**

### DIFF
--- a/docs/software_requirements/kernel_timing.sdoc
+++ b/docs/software_requirements/kernel_timing.sdoc
@@ -1,0 +1,231 @@
+[DOCUMENT]
+TITLE: Kernel Timing
+REQ_PREFIX: ZEP-SRS-28-
+
+[GRAMMAR]
+IMPORT_FROM_FILE: software_requirements.sgra
+
+[TEXT]
+STATEMENT: >>>
+SPDX-License-Identifier: Apache-2.0
+<<<
+
+[[SECTION]]
+TITLE: System Uptime
+
+[REQUIREMENT]
+UID: ZEP-SRS-28-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Timing
+TITLE: System uptime in milliseconds
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to obtain the time elapsed since the system booted, expressed in milliseconds.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-28
+
+[REQUIREMENT]
+UID: ZEP-SRS-28-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Timing
+TITLE: System uptime in system ticks
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to obtain the time elapsed since the system booted, expressed in system ticks.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-28
+
+[REQUIREMENT]
+UID: ZEP-SRS-28-3
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Timing
+TITLE: System uptime in seconds
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to obtain the time elapsed since the system booted, expressed in seconds.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-28
+
+[REQUIREMENT]
+UID: ZEP-SRS-28-4
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Timing
+TITLE: Elapsed time relative to a reference
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to obtain the time elapsed, in milliseconds, since a caller-provided reference time, and to replace that reference time with the current system uptime.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-28
+
+[[/SECTION]]
+
+[[SECTION]]
+TITLE: Hardware Clock
+
+[REQUIREMENT]
+UID: ZEP-SRS-28-5
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Timing
+TITLE: Hardware clock cycle count
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to read the current value of the system's hardware clock as a 32-bit cycle count.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-28
+
+[REQUIREMENT]
+UID: ZEP-SRS-28-6
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Timing
+TITLE: 64-bit hardware clock cycle count
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to read the current value of the system's hardware clock as a 64-bit cycle count.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-28
+
+[[/SECTION]]
+
+[[SECTION]]
+TITLE: Time Unit Conversion
+
+[REQUIREMENT]
+UID: ZEP-SRS-28-7
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Timing
+TITLE: Conversion between time units
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to convert time values between milliseconds, microseconds, nanoseconds, system ticks, and hardware cycles.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-28
+
+[[/SECTION]]
+
+[[SECTION]]
+TITLE: Thread Sleeping
+
+[REQUIREMENT]
+UID: ZEP-SRS-28-8
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Timing
+TITLE: Suspending a thread for a duration
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to suspend the execution of the calling thread for a specified duration, where the duration may be expressed in milliseconds, microseconds, nanoseconds, system ticks, or hardware cycles.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-34
+
+[REQUIREMENT]
+UID: ZEP-SRS-28-9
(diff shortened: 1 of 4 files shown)


docs/system_requirements/index.sdoc
```diff
@@ -171,6 +171,24 @@ RELATIONS:
...
+TITLE: Kernel Timing
+
+[REQUIREMENT]
+UID: ZEP-SYRS-28
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Timing
+TITLE: Kernel Timing
+STATEMENT: >>>
+The Zephyr RTOS shall provide services to measure the passage of time and to suspend the execution of a thread for a period of time.
```
nicpappler: This requirement is not atomic and needs to be split into two requirerements:
- one for the measuring
- one for the suspending of the execution of a thread

docs/software_requirements/kernel_timing.sdoc
```diff
@@ -0,0 +1,218 @@
...
+  VALUE: ZEP-SYRS-28
+
+[REQUIREMENT]
+UID: ZEP-SRS-28-4
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Timing
+TITLE: Elapsed time relative to a reference
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to obtain the time elapsed since a caller-provided reference time and to update that reference time to the current uptime.
```
nicpappler: This is not atomic and should be split into two requirements. As the two functionalities described are dependent on each other, we could argue that this "atomic" deviation can be accepted in favour of understandability. If we keep them together in one requirement, it is still not crear, what exactly is meant with "and update that reference time to the current uptime". Which time value will be overwritten with what time value?

docs/software_requirements/kernel_timing.sdoc
```diff
@@ -0,0 +1,218 @@
...
+TITLE: Suspending a thread for a duration
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to suspend the execution of the calling thread for a specified duration.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-28
+
+[REQUIREMENT]
+UID: ZEP-SRS-28-9
```
nicpappler, line 136: This requirement and the previous one (ZEP-SRS-28-8) seem to be either describing the same functionality (with ZEP-SRS-28-8 missing a time unit specification, or this ...28-9 requirement is a child of ...-28-8. I guess these two requirements need some more discussion or clarification.

docs/software_requirements/kernel_timing.sdoc
```diff
@@ -0,0 +1,218 @@
...
+  VALUE: ZEP-SYRS-28
+
+[REQUIREMENT]
+UID: ZEP-SRS-28-12
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Timing
+TITLE: Reporting remaining time on premature wake
+STATEMENT: >>>
+When a suspended thread is woken before its requested duration has elapsed, the Zephyr RTOS shall report the amount of time that was remaining.
```
nicpappler: Do we need a specific time unit here?

docs/software_requirements/kernel_timing.sdoc
```diff
@@ -0,0 +1,218 @@
...
+  VALUE: ZEP-SYRS-28
+
+[REQUIREMENT]
+UID: ZEP-SRS-28-13
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Timing
+TITLE: Busy waiting
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism for the calling thread to busy wait for a specified number of microseconds without relinquishing the CPU.
```
nicpappler, line 207: I don't see this as a child of ZEP-SYRS-28, we probably need another system level requirement for that.

docs/software_requirements/kernel_timing.sdoc
```diff
@@ -0,0 +1,218 @@
...
+TITLE: Timeouts
+
+[REQUIREMENT]
+UID: ZEP-SRS-28-14
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Kernel Timing
+TITLE: Absolute timeouts
+STATEMENT: >>>
+Where 64-bit timeouts are enabled, the Zephyr RTOS shall accept timeouts specified as an absolute point in time, in addition to timeouts specified as a relative duration.
```
nicpappler, line 225: I don't see this as a child of ZEP-SYRS-28, we probably need another system level requirement for that. Maybe also ZEP-SYRS-18 fits as a parent.

docs/software_requirements/timers.sdoc
```diff
@@ -206,3 +206,42 @@ When a timer expiry function is called, the Zephyr RTOS shall do so in the inter
...
+  VALUE: ZEP-SYRS-18
+
+[REQUIREMENT]
+UID: ZEP-SRS-4-17
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Timers
+TITLE: Timer cleanup with pending threads
+STATEMENT: >>>
+If a timer is cleaned up while one or more threads are still waiting on it, then the Zephyr RTOS shall return an error indicating that the cleanup could not be performed.
```
nicpappler: Does this mean that the timer also will not be cleaned up?

docs/software_requirements/timers.sdoc
```diff
@@ -206,3 +206,42 @@ When a timer expiry function is called, the Zephyr RTOS shall do so in the inter
...
+  VALUE: ZEP-SYRS-18
+
+[REQUIREMENT]
+UID: ZEP-SRS-4-18
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Timers
+TITLE: Observing timer lifecycle events
+STATEMENT: >>>
+Where timer observers are enabled, the Zephyr RTOS shall provide a mechanism to register observers that are notified of timer lifecycle events, including timer initialization, start, stop, and expiry.
```
nicpappler: Is this list complete?

nicpappler (changes requested):
There need to be some changes, and I also have some questions added for things that were not clear to me.

docs/software_requirements/timers.sdoc
nashif: It is complete: `struct k_timer_observer` under `CONFIG_TIMER_OBSERVER` has exactly four callbacks — `on_init`, `on_start`, `on_stop`, `on_expiry` — matching the four events listed.

text was changes:
```> ...notified of the following timer lifecycle events: initialization, start, stop, and expiry.```

docs/software_requirements/kernel_timing.sdoc
nashif: ZEP-SRS-28-8 was not missing a unit by accident: it refines the general sleep interface, which takes a duration expressed in any of the supported units. ZEP-SRS-28-9 refines the separate microsecond interface. What makes both necessary is not the unit of the requested duration but the unit of the *reported* one when a thread is woken early: the general mechanism reports milliseconds even when the duration was given in microseconds, while the microsecond interface reports microseconds.

made the change to name the units a duration may be expressed in for 28-8, and pairs 28-12/28-15 with the two reporting units

docs/software_requirements/kernel_timing.sdoc
nashif: Yes, and there is not a single one, which is why it was omitted. The general sleep mechanism reports the remaining time rounded up to the nearest millisecond; the microsecond interface (ZEP-SRS-28-9) rounds up to the nearest microsecond.

Split into two requirements:

- ZEP-SRS-28-12, remaining time expressed in milliseconds
- ZEP-SRS-28-15 (new), remaining time expressed in microseconds, the counterpart to ZEP-SRS-28-9

docs/software_requirements/kernel_timing.sdoc
nashif: Agreed. Busy waiting neither measures time nor suspends the thread — it explicitly keeps the CPU — so it fitted neither half of the original ZEP-SYRS-28.

Rather than add a fourth system requirement, made the change which phrases the second split as *delaying* the execution of a thread, which covers sleeping and busy waiting alike, and leaves the CPU-retention distinction at software level where it belongs. ZEP-SRS-28-13 now parents to ZEP-SYRS-34.

docs/software_requirements/kernel_timing.sdoc
nashif: Agreed that it does not belong under ZEP-SYRS-28. I would argue against ZEP-SYRS-18 though: that requirement is Timers, "a framework for managing time-based events", and absolute timeouts are not timer-specific. `K_TIMEOUT_ABS_TICKS/MS/US/NS/SEC/CYC` produce a timeout value accepted by every kernel API that takes one — `k_sleep()`, `k_sem_take()`, `k_queue_get()`, `k_timer_start()` and so on — so parenting it to Timers would understate its scope and make the traceability misleading.

added ZEP-SYRS-35 "Timeout specification" as the third split, and ZEP-SRS-28-14 now parents to it.

docs/software_requirements/timers.sdoc
nashif: Yes, nothing is released. `k_timer_cleanup()` returns `-EAGAIN` when threads are still pending on the timer's wait queue; the timer is left untouched and the caller keeps ownership of the storage, which it must not free.

The requirement stated only the error and left the post-condition open. now it is added:

> If a timer is cleaned up while one or more threads are still waiting on it, then the Zephyr RTOS shall leave the timer's resources unreleased and return an error indicating that the cleanup could not be performed.

docs/system_requirements/index.sdoc
nashif: Agreed, this was split. It ended up as three system requirements rather than two, because neither half of the original covered busy waiting or the way a timeout is specified:

- `ZEP-SYRS-28` Time measurement — parent of ZEP-SRS-28-1..28-7
- `ZEP-SYRS-34` Delaying thread execution — parent of ZEP-SRS-28-8..28-13
- `ZEP-SYRS-35` Timeout specification — parent of ZEP-SRS-28-14

That also resolves your two parentage comments before

docs/software_requirements/kernel_timing.sdoc
nashif: this is `k_uptime_delta()`, which returns `current_uptime - *reftime` in milliseconds and then overwrites the caller's `*reftime` with that same current uptime. So it is the caller's reference variable that is replaced, and it is replaced with the current system uptime. 
The two are inseparable in the interface, and splitting them would give a second requirement that can only be verified through the first — but made both halves explicit..

nicpappler (approved):
Looks good to merge to me.

PR #162, opened 2026-07-27
requirements: expand Logging requirements

nashif:
The Logging document covered 6 high-level requirements. Expand it to
26, reflecting the implemented and tested logging subsystem behavior
following the Route 3s approach: log message severity levels and
filtering (build-time and runtime), deferred and immediate processing
modes, log processing thread behavior, message dropping and flushing,
backend registration, activation and formatting, hexdump data logging,
printk redirection, and panic-mode synchronous flushing.

All requirements link to the logging system parent (ZEP-SYRS-8). UIDs
were generated with "strictdoc manage auto-uid".

Assisted-by: Claude:claude-fable-5

### DIFF
--- a/docs/software_requirements/logging.sdoc
+++ b/docs/software_requirements/logging.sdoc
@@ -105,3 +105,263 @@ As a Zephyr RTOS user I want a minimal influence of logging activities to the ti
 RELATIONS:
 - TYPE: Parent
   VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-7
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Severity levels
+STATEMENT: >>>
+The Zephyr RTOS logging shall support emitting log messages at multiple severity levels, including error, warning, informational, and debug.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-8
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Immediate logging support
+STATEMENT: >>>
+The Zephyr RTOS logging shall support immediate logging, in which a log message is processed in the context that generated it.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-9
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Compile-time severity filtering
+STATEMENT: >>>
+The Zephyr RTOS logging shall support a configurable compile-time severity level below which log messages are removed from the build.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-10
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Runtime severity filtering
+STATEMENT: >>>
+The Zephyr RTOS logging shall support adjusting the severity filtering level at run time, per source and per backend.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-11
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Log source registration
+STATEMENT: >>>
+The Zephyr RTOS logging shall allow software modules and instances to register as independent log sources.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-12
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Binary data logging
+STATEMENT: >>>
+The Zephyr RTOS logging shall support logging a block of binary data as a hexadecimal dump.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-13
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Flushing pending log messages
+STATEMENT: >>>
+The Zephyr RTOS logging shall provide a mechanism to flush pending log messages, including before the system halts on a fatal error.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-14
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Message timestamping
+STATEMENT: >>>
+The Zephyr RTOS logging shall attach a timestamp to each log message at the time the message is generated.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-15
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Custom timestamp source
+STATEMENT: >>>
+The Zephyr RTOS logging shall provide a mechanism to configure a custom function and frequency used to obtain the timestamp for log messages.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-16
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Alternative output formats
+STATEMENT: >>>
+The Zephyr RTOS logging shall support rendering log message output in alternative machine-readable formats, including a MIPI SyS-T format and a dictionary-based format.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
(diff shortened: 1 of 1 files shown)


docs/software_requirements/logging.sdoc
```diff
@@ -105,3 +105,263 @@ As a Zephyr RTOS user I want a minimal influence of logging activities to the ti
...
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-9
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Compile-time severity filtering
+STATEMENT: >>>
+The Zephyr RTOS logging shall support a configurable compile-time severity level below which log messages are removed from the build.
```
nicpappler: This requirement has some weak phrasing. "below" is ambiguous, I assume it means something like "lower level than ???", but there is room for interpretation right now. 
Also it is not sure how something can be "removed from the build" - is it removed after something has been build (technically I don't think so), or will it be excluded from being build at all? I can also just guess that this relates also to the compile-time as stated in the beginning of this requirement, but there is room for interpretation.

docs/software_requirements/logging.sdoc
```diff
@@ -105,3 +105,263 @@ As a Zephyr RTOS user I want a minimal influence of logging activities to the ti
...
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-13
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Flushing pending log messages
+STATEMENT: >>>
+The Zephyr RTOS logging shall provide a mechanism to flush pending log messages, including before the system halts on a fatal error.
```
nicpappler: Just as a comment (and maybe to be discussed with @tobiaskaestner and @parphane ). This should formally be two requirements, but from an understandability perspective, I would like to keep this in one requirement, consciously accepting the violation of the atomic need.

docs/software_requirements/logging.sdoc
```diff
@@ -105,3 +105,263 @@ As a Zephyr RTOS user I want a minimal influence of logging activities to the ti
...
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-16
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Alternative output formats
+STATEMENT: >>>
+The Zephyr RTOS logging shall support rendering log message output in alternative machine-readable formats, including a MIPI SyS-T format and a dictionary-based format.
```
nicpappler: Are these these the only two formats? We need to make sure we are listing all possible formats (or refer to a design decision that will define this for us "later" in the traceability chain).

docs/software_requirements/logging.sdoc
```diff
@@ -105,3 +105,263 @@ As a Zephyr RTOS user I want a minimal influence of logging activities to the ti
...
   VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-7
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Severity levels
+STATEMENT: >>>
+The Zephyr RTOS logging shall support emitting log messages at multiple severity levels, including error, warning, informational, and debug.
```
nicpappler, line 116: Are these all severity levels? To be unambiguous, we need to either list all, or define this in a design document later in the traceability chain.

nicpappler (changes requested):
Formally we need a few changes to remove ambiguities.

docs/software_requirements/logging.sdoc
tobiaskaestner: I read the list here as minimally required but not excluding further log levels. That's indeed something for the design specification

docs/software_requirements/logging.sdoc
```diff
@@ -105,3 +105,263 @@ As a Zephyr RTOS user I want a minimal influence of logging activities to the ti
...
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-10
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Runtime severity filtering
+STATEMENT: >>>
+The Zephyr RTOS logging shall support adjusting the severity filtering level at run time, per source and per backend.
```
tobiaskaestner, line 155: A requirement, that Zephyr shall support different logging backends at all is missing

docs/software_requirements/logging.sdoc
```diff
@@ -105,3 +105,263 @@ As a Zephyr RTOS user I want a minimal influence of logging activities to the ti
...
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-11
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Log source registration
+STATEMENT: >>>
+The Zephyr RTOS logging shall allow software modules and instances to register as independent log sources.
```
tobiaskaestner: Not clear what instances is referring to here.

docs/software_requirements/logging.sdoc
tobiaskaestner: Agree, this is two requirements. A general mechanisms to flush on user request. And a second to garantuee this mechanism is used to flush in case of a fatal error.

docs/software_requirements/logging.sdoc
tobiaskaestner: Agreed, sufficient to say that machine-readable formats need to be supported, but leave what's actually supported to the design specification

docs/software_requirements/logging.sdoc
```diff
@@ -105,3 +105,263 @@ As a Zephyr RTOS user I want a minimal influence of logging activities to the ti
...
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-18
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Custom output format
+STATEMENT: >>>
+Where a custom output format is enabled, the Zephyr RTOS logging shall use an application-provided function to format log message output.
```
tobiaskaestner: The term custom is vague here. If I recall correctly, there is a specific entry in Kconfig called Custom, so users can select Custom to provide their own format log function. 
In simpler terms this requirement says, that Users can configure Zephyr to use a custom log format function. That they then have to provide a handler actually becomes an obligation on their behalf then.

docs/software_requirements/logging.sdoc
```diff
@@ -105,3 +105,263 @@ As a Zephyr RTOS user I want a minimal influence of logging activities to the ti
...
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-20
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Routing printk output through logging
+STATEMENT: >>>
+Where printk integration is enabled, the Zephyr RTOS shall route printk output through the logging subsystem.
```
tobiaskaestner: If we want this requirement to be included here, then a requirement what printk actually is, is required as well. Otherwise the requirements are incomplete and thus inconsistent.

I'd opt to exlclude this as a requirement for now.

docs/software_requirements/logging.sdoc
```diff
@@ -105,3 +105,263 @@ As a Zephyr RTOS user I want a minimal influence of logging activities to the ti
...
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-22
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Dropped message accounting on buffer overflow
+STATEMENT: >>>
+While operating in deferred mode, if the log message buffer is full, then the Zephyr RTOS logging shall drop log messages and account for the number of messages dropped.
```
tobiaskaestner: not clear if oldest or newest messages are dropped.

docs/software_requirements/logging.sdoc
```diff
@@ -105,3 +105,263 @@ As a Zephyr RTOS user I want a minimal influence of logging activities to the ti
...
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-24
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Multi-domain log aggregation
+STATEMENT: >>>
+Where multi-domain logging is enabled, the Zephyr RTOS logging shall aggregate log messages originating from multiple domains, including remote domains.
```
tobiaskaestner, line 371: Where are multi-domains and remote domains defined in the requirements spec? Without saying what that is, this requirement is incomplete.

docs/software_requirements/logging.sdoc
```diff
@@ -105,3 +105,263 @@ As a Zephyr RTOS user I want a minimal influence of logging activities to the ti
...
+  VALUE: ZEP-SYRS-8
+
+[REQUIREMENT]
+UID: ZEP-SRS-11-26
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Logging
+TITLE: Time-ordered processing of linked-domain messages
+STATEMENT: >>>
+Where multi-domain logging with dedicated link buffers is enabled, the Zephyr RTOS logging shall process messages received from log links in timestamp order.
```
tobiaskaestner: Where are link buffers required? Again, same incompleteness argument as above.

tobiaskaestner (changes requested):
@nashif please take another look. 

The PR introduces a couple of technical terms without defining or requiring them. This is affecting the self-consistency of the requirements spec.

nashif:
reworded/split as suggested

docs/software_requirements/logging.sdoc
nashif: Covered by ZEP-SRS-11-5, which was easy to miss under its old wording — reworded to name logging backends explicitly

tobiaskaestner (approved):
LGTM, all points of my previous review have been addressed,
many thanks @nashif

PR #163, opened 2026-07-27
requirements: expand Memory Objects requirements

nashif:
The Memory Objects document covered only 2 requirements. Expand it to
18, reflecting the implemented and tested kernel memory services
following the Route 3s approach: memory heap definition and
run-time initialization, allocation with alignment and timeout,
release from any context, memory slabs (definition, allocation,
release, usage statistics), the system heap, and enumerating defined
heaps.

All requirements link to the memory objects system parent
(ZEP-SYRS-9). UIDs were generated with "strictdoc manage auto-uid".

Assisted-by: Claude:claude-fable-5


nicpappler (approved):
Looks formally ok to me. If there are no other findings regarding completeness from others, it is good to merge.

PR #164, opened 2026-07-27
Memory Protection + Virtual Memory

nashif:
- **requirements: expand Memory Protection requirements**
- **requirements: add Virtual Memory software requirements**

### DIFF
--- a/docs/software_requirements/memory_protection.sdoc
+++ b/docs/software_requirements/memory_protection.sdoc
@@ -10,6 +10,13 @@ STATEMENT: >>>
 SPDX-License-Identifier: Apache-2.0
 <<<
 
+[TEXT]
+STATEMENT: >>>
+Unless a requirement states otherwise, the requirements in this document
+apply where user mode support is enabled. The stack canary requirements
+and the write-XOR-execute enforcement do not depend on user mode support.
+<<<
+
 [REQUIREMENT]
 UID: ZEP-SRS-8-1
 STATUS: Draft
@@ -174,7 +181,7 @@ TYPE: Functional
 COMPONENT: Memory Protection
 TITLE: User mode handling of detected stack overflow
 STATEMENT: >>>
-The Zephyr RTOS shall support a defined mechanism for user mode handling a of detected stack overflow.
+The Zephyr RTOS shall support a defined mechanism for user mode handling of a detected stack overflow.
 <<<
 USER_STORY: >>>
 As a Zephyr RTOS user I want, when a stack overflow is detected, to be able to implement a graceful, application defined handling of the exception.
@@ -222,7 +229,7 @@ STATEMENT: >>>
 The Zephyr RTOS shall provide helper functions for system call handler functions to validate the inputs passed in from user mode before invoking the implementation function to protect the kernel.
 <<<
 USER_STORY: >>>
-As a Zephyr RTOS user I want Zepyhr OS to validate system call parameters passed from the user mode to the kernel mode to avoid crashes and undefined behaviour.
+As a Zephyr RTOS user I want Zephyr OS to validate system call parameters passed from the user mode to the kernel mode to avoid crashes and undefined behaviour.
 <<<
 RELATIONS:
 - TYPE: Parent
@@ -238,12 +245,7 @@ STATEMENT: >>>
 The Zephyr RTOS shall support system calls to be able to safely accept C strings passed in from user mode.
 <<<
 USER_STORY: >>>
-As a Zephyr RTOS user I want Zepyhr OS to validate system call string type parameters passed from the user mode to the kernel mode to avoid crashes and undefined behaviour.
-
-e.g.
-
-- verify the string length is smaller or equal to the syscalls defined max.
-- verify that the length type does not overflow when allocating one more byte ???
+As a Zephyr RTOS user I want Zephyr OS to validate system call string type parameters passed from the user mode to the kernel mode to avoid crashes and undefined behaviour, for example verifying that the string length does not exceed the maximum defined for the system call.
 <<<
 RELATIONS:
 - TYPE: Parent
@@ -254,11 +256,9 @@ UID: ZEP-SRS-8-16
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Memory Protection
-TITLE: Tracking kernel objects in used by user mode threads
+TITLE: Tracking kernel objects in use by user mode threads
 STATEMENT: >>>
-The Zephyr RTOS shall track kernel objects that are used by user mode threads.
-
-Note: this means Zephyr shall track the resources used by the user mode thread (associate this with a user story).
+The Zephyr RTOS shall track the kernel objects that are in use by each user mode thread.
 <<<
 RELATIONS:
 - TYPE: Parent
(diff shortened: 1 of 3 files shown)


docs/software_requirements/memory_protection.sdoc
```diff
@@ -295,3 +288,120 @@ As a Zephyr RTOS user I want to be able, during runtime from the kernel, to requ
...
+  VALUE: ZEP-SYRS-9
+
+[REQUIREMENT]
+UID: ZEP-SRS-8-24
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Memory Protection
+TITLE: Default user thread memory isolation
+STATEMENT: >>>
+While a thread runs in user mode, the Zephyr RTOS shall restrict its memory accesses to its own stack and the partitions of the memory domain to which it is assigned.
```
JoeyFreeland-DojoFive, line 364: How will this affect the way application user threads interact with system threads or other system services? For example, will a user thread be able to send messages to a system service's queue (ie, network manager). What about system modules that expose functions to modify its internal static data? Maybe some more context around the partitioning system as a whole would help, and how granular you intend that partitioning to be.

docs/software_requirements/memory_protection.sdoc
```diff
@@ -295,3 +288,120 @@ As a Zephyr RTOS user I want to be able, during runtime from the kernel, to requ
...
+  VALUE: ZEP-SYRS-9
+
+[REQUIREMENT]
+UID: ZEP-SRS-8-20
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Memory Protection
+TITLE: Memory domain definition
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to define and initialize a memory domain composed of a set of memory partitions.
```
anobli: I have found no macro that defines and initialize a memory domain. There is only one function for initializing them so I think we should remove define from requirement.

docs/software_requirements/memory_protection.sdoc
```diff
@@ -295,3 +288,120 @@ As a Zephyr RTOS user I want to be able, during runtime from the kernel, to requ
...
+  VALUE: ZEP-SYRS-9
+
+[REQUIREMENT]
+UID: ZEP-SRS-8-21
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Memory Protection
+TITLE: Adding a partition to a memory domain
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to add a memory partition to a memory domain.
```
anobli: Maybe it is worth to mention this is for an initialized domain.

docs/software_requirements/memory_protection.sdoc
```diff
@@ -295,3 +288,120 @@ As a Zephyr RTOS user I want to be able, during runtime from the kernel, to requ
...
+<<<
```
anobli, line 359: Both related to init and add partition: we miss requirement for memory constraints check by both functions:
- If CONFIG_EXECUTE_XOR_WRITE is enabled, the partition must not allow both writes and execution.
- Partitions in the same memory domain may not overlap each other.

There are other constraints in documentation but if they are checked by the code, this is done by hardware specific code so I am not sure this is in our scope.

docs/software_requirements/memory_protection.sdoc
```diff
@@ -295,3 +288,120 @@ As a Zephyr RTOS user I want to be able, during runtime from the kernel, to requ
...
+  VALUE: ZEP-SYRS-9
+
+[REQUIREMENT]
+UID: ZEP-SRS-8-23
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Memory Protection
+TITLE: Assigning a thread to a memory domain
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to assign a thread to a memory domain so that the thread gains access to the domain's partitions.
```
anobli, line 384: We need a second requirement to explain that if the thread was already assigned to a memory domain, it get removed from that memory domain.

docs/software_requirements/memory_protection.sdoc
```diff
@@ -295,3 +288,120 @@ As a Zephyr RTOS user I want to be able, during runtime from the kernel, to requ
...
+  VALUE: ZEP-SYRS-9
+
+[REQUIREMENT]
+UID: ZEP-SRS-8-25
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Memory Protection
+TITLE: Stack canary protection
+STATEMENT: >>>
+Where stack canary protection is enabled, the Zephyr RTOS shall detect corruption of a function's stack frame using a guard value and treat the corruption as a fatal error.
```
anobli: I guess that we need two requirements here: one to define that where stack canary protection is enabled, we should detect the corruption and one to describe behavior when this happens: ` If stack frame corruption is detected ...`

docs/software_requirements/virtual_memory.sdoc
```diff
@@ -0,0 +1,216 @@
...
+TITLE: Virtual Memory Mapping
+
+[REQUIREMENT]
+UID: ZEP-SRS-36-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Virtual Memory
+TITLE: Mapping physical memory into the virtual address space
+STATEMENT: >>>
+Where virtual memory is enabled, the Zephyr RTOS shall provide a mechanism to map a region of physical memory into the virtual address space with caller-specified access permissions.
```
anobli: Is this requirement for `k_mem_map_phys_guard`?
It looks like this is an internal kernel API, do we really need a requirement for it ?

docs/software_requirements/virtual_memory.sdoc
```diff
@@ -0,0 +1,216 @@
...
+  VALUE: ZEP-SYRS-9
+
+[REQUIREMENT]
+UID: ZEP-SRS-36-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Virtual Memory
+TITLE: Unmapping a virtual memory region
+STATEMENT: >>>
+Where virtual memory is enabled, the Zephyr RTOS shall provide a mechanism to unmap a previously mapped virtual memory region so that subsequent accesses to it are no longer permitted.
```
anobli: Same here!

docs/software_requirements/virtual_memory.sdoc
```diff
@@ -0,0 +1,216 @@
...
+  VALUE: ZEP-SYRS-9
+
+[REQUIREMENT]
+UID: ZEP-SRS-36-4
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Virtual Memory
+TITLE: Anonymous memory mapping
+STATEMENT: >>>
+Where virtual memory is enabled, the Zephyr RTOS shall provide a mechanism to allocate and map an anonymous region of memory and to release it back to the system.
```
anobli: We need two requirements, one for allocating and mapping memory, and one for releasing the resources back to the system.

docs/software_requirements/virtual_memory.sdoc
```diff
@@ -0,0 +1,216 @@
...
+  VALUE: ZEP-SYRS-9
+
+[REQUIREMENT]
+UID: ZEP-SRS-36-5
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Virtual Memory
+TITLE: Guard pages for anonymous memory mappings
+STATEMENT: >>>
+When an anonymous memory region is mapped, the Zephyr RTOS shall place inaccessible guard pages adjacent to the region so that accesses beyond its bounds are detected.
```
anobli: Please use before and after instead of adjacent. This is less ambiguous.

docs/software_requirements/virtual_memory.sdoc
```diff
@@ -0,0 +1,216 @@
...
+  VALUE: ZEP-SYRS-9
+
+[REQUIREMENT]
+UID: ZEP-SRS-36-6
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Virtual Memory
+TITLE: User-mode-accessible memory mappings
+STATEMENT: >>>
+Where user mode is enabled, the Zephyr RTOS shall provide a mechanism to map a memory region that is accessible from user mode threads.
```
anobli: What API this requirement is describing ?

anobli (commented):
I commented some requirements. I still have to review everything related to demand paging.
Also, I have a concern regarding the memory protection requirements since they all depends on  USER being enabled. This is not mentioned anywhere but in meanwhile, it would be cumbersome adding the where statement anywhere.

docs/software_requirements/virtual_memory.sdoc
nashif: the K_MEM_PERM_USER mapping flag, verified by test_k_mem_map_user

docs/software_requirements/memory_protection.sdoc
nashif: User threads reach kernel objects and services only through system calls on objects they've been granted access to (the object-access-grant requirements cover that); partition granularity is MPU/MMU-region-defined per architecture.

docs/software_requirements/memory_protection.sdoc
```diff
@@ -295,3 +295,172 @@ As a Zephyr RTOS user I want to be able, during runtime from the kernel, to requ
...
+  VALUE: ZEP-SYRS-9
+
+[REQUIREMENT]
+UID: ZEP-SRS-8-24
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Memory Protection
+TITLE: Default user thread memory isolation
+STATEMENT: >>>
+While a thread runs in user mode, the Zephyr RTOS shall restrict its memory accesses to its own stack and the partitions of the memory domain to which it is assigned.
```
anobli, line 410: I think we need a requirement to explain what is the system response if a user thread tries to read or write in memory of another thread or memory partition.

docs/software_requirements/virtual_memory.sdoc
```diff
@@ -0,0 +1,216 @@
...
+  VALUE: ZEP-SYRS-9
+
+[REQUIREMENT]
+UID: ZEP-SRS-36-6
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Virtual Memory
+TITLE: User-mode-accessible memory mappings
+STATEMENT: >>>
+Where user mode is enabled, the Zephyr RTOS shall control, for each memory mapping, whether user mode threads may access the mapping.
```
anobli, line 88: This does not really explain what the system must do.
As example, what happens when a thread try accessing a mapped memory that it doesn't have right to access to.

docs/software_requirements/virtual_memory.sdoc
```diff
@@ -0,0 +1,216 @@
...
+  VALUE: ZEP-SYRS-9
+
+[REQUIREMENT]
+UID: ZEP-SRS-36-13
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Virtual Memory
+TITLE: Collecting demand paging statistics
+STATEMENT: >>>
+Where demand paging statistics are enabled, the Zephyr RTOS shall provide a mechanism to query system-wide and per-thread demand paging statistics.
```
anobli, line 184: We have two different API, one for system, one for thread so we need two requirements.

docs/software_requirements/virtual_memory.sdoc
```diff
@@ -0,0 +1,216 @@
...
+  VALUE: ZEP-SYRS-9
+
+[REQUIREMENT]
+UID: ZEP-SRS-36-14
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Virtual Memory
+TITLE: Collecting demand paging timing histograms
+STATEMENT: >>>
+Where demand paging timing histograms are enabled, the Zephyr RTOS shall provide a mechanism to query histograms of page eviction and backing store access times.
```
anobli, line 197: Same here, we need two requirements, one for page eviction and one for backing store.

docs/software_requirements/virtual_memory.sdoc
```diff
@@ -0,0 +1,216 @@
...
+  VALUE: ZEP-SYRS-9
+
+[REQUIREMENT]
+UID: ZEP-SRS-36-15
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Virtual Memory
+TITLE: On-demand evictable code and data sections
+STATEMENT: >>>
+Where demand paging is enabled, the Zephyr RTOS shall support placing code and data in sections that are loaded into physical memory on demand and may be evicted when not in use.
```
anobli, line 210: Is not section implementation detail ?

PR #159, opened 2026-07-25
reorder indexes alphabetically

nashif:
just reordering, to avoild conflicts with too many PRs inserting items in the
same spot.

- **requirements: sort software requirements index alphabetically**
- **requirements: sort system requirements index alphabetically**

PR #168, opened 2026-08-26
Add spinlock service requirements

gilsken:
This PR adds the software requirements for the spinlock service.


nashif:
- REQ_PREFIX: ZEP-SRS-36- is already claimed by open PR #164. Please take the next free document number — 39 is not claimed by any open PR or pending set 
- Both indexes are alphabetically sorted since #159. spinlock.sdoc belongs between semaphore.sdoc and stacks.sdoc, not appended after tracing.sdoc
- ZEP-SRS-2-3 "Exclusion between physical CPUs" already exists on main (moving to ZEP-SRS-34-3 in #142)

docs/software_requirements/spinlock.sdoc
```diff
@@ -0,0 +1,158 @@
...
+<<<
+
+[REQUIREMENT]
+UID: ZEP-SRS-36-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Spinlock Service
+TITLE: Spinlock Key Opaque Type
+STATEMENT: >>>
+The Zephyr RTOS shall define an opaque type. This type shall encapsulate the architecture-dependent CPU interrupt state at the time of spinlock acquisition.
```
nashif: not behavioral requirements — they describe data types (VerReq_1_5). Drop; the key concept is captured by stating that acquisition returns a key capturing the local interrupt state and release takes it back.

docs/software_requirements/spinlock.sdoc
```diff
@@ -0,0 +1,158 @@
...
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-36
+
+[REQUIREMENT]
+UID: ZEP-SRS-36-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Spinlock Service
+TITLE: Spinlock struct
```
nashif: ditto

docs/software_requirements/spinlock.sdoc
```diff
@@ -0,0 +1,158 @@
...
+[REQUIREMENT]
+UID: ZEP-SRS-36-3
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Spinlock Service
+TITLE: Synchronous Spinlock Acquisition
+STATEMENT: >>>
+The Zephyr RTOS shall provide an interface to lock a spinlock struct. Upon invocation, this interface shall guarantee that the calling thread is neither suspended nor interrupted on its current CPU until a matching unlock is performed, and shall guarantee that exactly one thread on one CPU holds the lock at a time.
+A spinlock key is returned as handle that shall be provided to unlock API to restore the local CPU's interrupt state.
+Holding a spinlock while a context switch occurs is illegal. Attempting to lock a spinlock struct that is already held by the calling CPU shall result in a deadlock, as spinlocks are not recursive.
```
nashif: bundles ≥5 things (mutual exclusion, no suspension/interruption, key handle, "illegal while context switching", non-recursive deadlock) — not atomic (VerReq_1_1). "Shall result in a deadlock" and "is illegal" are caller constraints, not verifiable behavior; either drop them or express them as the validation-framework detections in 36-10

docs/software_requirements/spinlock.sdoc
```diff
@@ -0,0 +1,158 @@
...
+  VALUE: ZEP-SYRS-36
+
+[REQUIREMENT]
+UID: ZEP-SRS-36-4
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Spinlock Service
+TITLE: Synchronous Spinlock Acquisition in SMP
+STATEMENT: >>>
+Where the SMP is enabled, the Zephyr RTOS shall continuously spin-poll the lock's internal atomic state until it is successfully acquired, guaranteeing mutual exclusion between the CPUs contending for the same spinlock struct.
```
nashif: "spin-poll the lock's internal atomic state" is implementation. Also parented to an SRS while restating 36-3's exclusion clause. Suggested replacement (implementation-free): "While a spinlock is held by another CPU, when an execution context attempts to acquire it, the Zephyr RTOS shall delay the attempt until the spinlock is released."

docs/software_requirements/spinlock.sdoc
```diff
@@ -0,0 +1,158 @@
...
+- TYPE: Parent
+  VALUE: ZEP-SRS-36-3
+
+[REQUIREMENT]
+UID: ZEP-SRS-36-5
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Spinlock Service
+TITLE: Non-Blocking Spinlock Attempt
+STATEMENT: >>>
```
nashif, line 71: mixes the mechanism, interrupt handling and return-value protocol ("value of 0", "-EBUSY", "output parameter"). Split into the mechanism and the busy case: "If a non-blocking acquisition attempt finds the spinlock held, then the Zephyr RTOS shall leave the local CPU's interrupt state unchanged and return an error indicating the spinlock is busy."

docs/software_requirements/spinlock.sdoc
```diff
@@ -0,0 +1,158 @@
...
+  VALUE: ZEP-SYRS-36
+
+[REQUIREMENT]
+UID: ZEP-SRS-36-6
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Spinlock Service
+TITLE: Standard Spinlock Unlocking
+STATEMENT: >>>
+The Zephyr RTOS shall provide an unlock interface to release atomically a locked spinlock struct, making it available to other execution contexts and restore the previous CPU's interrupt state with the key opaque type.
```
nashif: three concerns (release+restore, nesting order, list of illegal uses). Keep the release/restore statement; state nesting positively as its own requirement (separate spinlocks may be acquired while holding another); the "illegal" list again belongs to the validation detections.

docs/software_requirements/spinlock.sdoc
```diff
@@ -0,0 +1,158 @@
...
+  VALUE: ZEP-SRS-36-7
+
+[REQUIREMENT]
+UID: ZEP-SRS-36-9
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Spinlock Service
+TITLE: Scope-Exit Spinlock Misuse Detection
+STATEMENT: >>>
+Where the spinlock validation framework is enabled and the toolchain supports scope-exit cleanup attributes, the Zephyr RTOS shall detect, by means of a runtime assertion, any exit from a K_SPINLOCK-guarded code block that is not performed through reaching the end of the block or through the K_SPINLOCK_BREAK macro (e.g. exiting the block with a break, goto or return statement).
```
nashif: K_SPINLOCK, K_SPINLOCK_BREAK, "toolchain supports scope-exit cleanup attributes" are API and toolchain specifics. Describe the behavior — "a mechanism to execute a block of code while holding a spinlock, releasing it when the block completes or is exited early through the mechanism's early-exit facility"

docs/software_requirements/spinlock.sdoc
```diff
@@ -0,0 +1,158 @@
...
+  VALUE: ZEP-SRS-36-7
+
+[REQUIREMENT]
+UID: ZEP-SRS-36-10
+STATUS: Draft
+TYPE: Non-Functional
+COMPONENT: Spinlock Service
+TITLE: Spinlock Validation Framework
+STATEMENT: >>>
+Where the spinlock validation framework is enabled, the Zephyr RTOS shall detect at runtime, by means of an assertion, an attempt to lock an invalid spinlock struct, or an attempt to unlock a spinlock struct that is not held by the calling context.
```
nashif: "invalid spinlock struct" is vague (VerReq_1_0). What the framework actually detects, and what the tests verify, is precise — recommend three requirements under Where spinlock validation is enabled: (a) acquisition of a spinlock already held by the calling CPU, (b) release of a spinlock not held by the caller, (c) a context switch while a spinlock is held. 36-11 is fine as the fourth (hold-time limit); note it also depends on a lock-free cycle counter — leave that to Kconfig.

nashif (changes requested):

docs/system_requirements/index.sdoc
```diff
@@ -523,3 +523,16 @@ As a Zephyr RTOS user I want an ISR or a high-priority thread to defer non-urgen
...
+UID: ZEP-SYRS-39
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Spinlock
+TITLE: Spinlock
+STATEMENT: >>>
+The Zephyr RTOS shall provide a low-level spinlock synchronization mechanism to protect critical sections across multi-core systems while managing local CPU interrupt states.
+<<<
+USER_STORY: >>>
+As a Zephyr RTOS user, I want to be able to lock and unlock specific code in a multi-core systems.
```
anobli, line 537: I find it a bit ambiguous. I would have say something like that as user, I want be able to protect code execution against concurrent execution in a multi-core systems.

docs/system_requirements/index.sdoc
```diff
@@ -523,3 +523,16 @@ As a Zephyr RTOS user I want an ISR or a high-priority thread to defer non-urgen
 <<<
 
 [[/SECTION]]
+
+[REQUIREMENT]
+UID: ZEP-SYRS-39
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Spinlock
```
anobli, line 531: Inconsistency with COMPONENT value used everywhere in spinlock.sdoc. You should use `Spinlock Service` here or only use `Spinlock everywhere`.

docs/software_requirements/spinlock.sdoc
```diff
@@ -0,0 +1,219 @@
...
+<<<
+
+[REQUIREMENT]
+UID: ZEP-SRS-39-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Spinlock Service
+TITLE: Spinlock Acquisition
+STATEMENT: >>>
+The spinlock acquire operation shall lock the specified spinlock and return a key that captures the local interrupt state prior to acquisition.
```
anobli, line 20: Inconsistency issue here. The requirement should look like `The Zephyr RTOS shall provide a mechanism to lock a ...`

docs/software_requirements/spinlock.sdoc
```diff
@@ -0,0 +1,219 @@
...
+  VALUE: ZEP-SYRS-39
+
+[REQUIREMENT]
+UID: ZEP-SRS-39-2
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Spinlock Service
+TITLE: Spinlock Release
+STATEMENT: >>>
+The spinlock release operation shall take the key returned by the acquisition operation, atomically unlock the specified spinlock, and restore the local interrupt state captured in the key.
```
anobli, line 33: Same inconsistency issue here.

docs/software_requirements/spinlock.sdoc
```diff
@@ -0,0 +1,219 @@
...
+  VALUE: ZEP-SYRS-39
+
+[REQUIREMENT]
+UID: ZEP-SRS-39-3
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Spinlock Service
+TITLE: Spinlock Mutual Exclusion
+STATEMENT: >>>
+Upon successful acquisition of a spinlock, the Zephyr RTOS shall guarantee that exactly one execution context holds the lock at any given time.
```
anobli, line 46: This doesn't respect EARS. I would suggest: `When a spinlock is successfully acquired, `

docs/software_requirements/spinlock.sdoc
```diff
@@ -0,0 +1,219 @@
...
+  VALUE: ZEP-SYRS-39
+
+[REQUIREMENT]
+UID: ZEP-SRS-39-4
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Spinlock Service
+TITLE: Spinlock CPU State Protection
+STATEMENT: >>>
+While a spinlock is held, the Zephyr RTOS shall prevent the calling thread from being suspended or interrupted on its current CPU.
```
anobli, line 59: This doesn't mention ISR, only thread, whereas this should also apply to ISR. Here, I would suggest to keep using execution context.

docs/software_requirements/spinlock.sdoc
```diff
@@ -0,0 +1,219 @@
...
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-39
+
+[REQUIREMENT]
+UID: ZEP-SRS-39-5
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Spinlock Service
+TITLE: Synchronous Spinlock Acquisition in SMP
```
anobli, line 70: Synchronous word is weird here. Maybe blocking would be better, especially if we consider that later we talk about use non-blocking words.

docs/software_requirements/spinlock.sdoc
```diff
@@ -0,0 +1,219 @@
...
+STATEMENT: >>>
+While a spinlock is held by another CPU, when an execution context attempts to acquire it, the Zephyr RTOS shall delay the attempt until the spinlock is released.
```
anobli, line 72: I don't like delay word in this context. Maybe busy-wait is more accurate ?

anobli: Also, this requirement doesn't what happens when multiple CPUs wait for the same lock ?
Which one is going to take the lock first once it has been released ?
Zephyr has two different way to manage it, so i guess we would need a second requirement for that.

docs/software_requirements/spinlock.sdoc
```diff
@@ -0,0 +1,219 @@
...
+  VALUE: ZEP-SYRS-39
+
+[REQUIREMENT]
+UID: ZEP-SRS-39-6
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Spinlock Service
+TITLE: Non-Blocking Spinlock Attempt Success
+STATEMENT: >>>
+Upon a non-blocking acquisition attempt where the spinlock is available, the Zephyr RTOS shall acquire the lock without spinning and return a key capturing the local interrupt state prior to acquisition.
```
anobli, line 85: EARS is not respected here.  The requirement should start whith `When a non-blocking acquisition attempt  ...`

docs/software_requirements/spinlock.sdoc
```diff
@@ -0,0 +1,219 @@
...
+  VALUE: ZEP-SYRS-39
+
+[REQUIREMENT]
+UID: ZEP-SRS-39-8
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Spinlock Service
+TITLE: Spinlock Nesting Capability
+STATEMENT: >>>
+The Zephyr RTOS shall allow a thread to acquire separate spinlocks while already holding one or more spinlocks, provided they are released in the strictly reverse order of their acquisition.
```
anobli, line 111: Please use context execution instead of thread

docs/software_requirements/spinlock.sdoc
```diff
@@ -0,0 +1,219 @@
...
+  VALUE: ZEP-SYRS-39
+
+[REQUIREMENT]
+UID: ZEP-SRS-39-9
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Spinlock Service
+TITLE: Scoped Spinlock Execution
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to execute a block of code while holding a spinlock, automatically releasing the spinlock and restoring the prior interrupt state when the block completes or is exited early through the mechanism's provided early-exit facility.
```
anobli, line 124: This requirement is not atomic. we should have have for the locking mechanism and one for the early exit.

anobli (changes requested):

PR #169, opened 2026-09-06
requirements: expand Exception and Error Handling requirements

nashif:
The document covered default and assignable handlers but not triggering
fatal errors or the handler contract. Add four requirements
(ZEP-SRS-16-5 .. ZEP-SRS-16-8): triggering a fatal error that terminates
the current context (kernel oops), triggering a fatal error that
terminates the system (kernel panic), providing a reason code to the
fatal error handler, and the default fatal error policy (terminate the
offending thread in thread context, otherwise halt the system).

The requirements reflect the implemented k_oops, k_panic, k_fatal_halt,
and k_sys_fatal_error_handler behavior following the Route 3s approach
and link to the system parent (ZEP-SYRS-5). UIDs were generated with
"strictdoc manage auto-uid".

### DIFF
--- a/docs/software_requirements/exception_and_error_handling.sdoc
+++ b/docs/software_requirements/exception_and_error_handling.sdoc
@@ -64,3 +64,55 @@ The Zephyr RTOS shall provide an interface to assign a specific handler for a fa
 RELATIONS:
 - TYPE: Parent
   VALUE: ZEP-SYRS-5
+
+[REQUIREMENT]
+UID: ZEP-SRS-16-5
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Exception and Error Handling
+TITLE: Triggering a fatal error in the current context
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism for code to trigger a fatal error that terminates the currently running context.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-5
+
+[REQUIREMENT]
+UID: ZEP-SRS-16-6
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Exception and Error Handling
+TITLE: Triggering a fatal system error
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism for code to trigger a fatal error that terminates the system.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-5
+
+[REQUIREMENT]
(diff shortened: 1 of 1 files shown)


docs/software_requirements/exception_and_error_handling.sdoc
anobli, line 49: I think requirement is incorrect: we don't have any API for that, just a weak function that we could override.

anobli, line 62: I think requirement is incorrect: we don't have any API for that, just a weak function that we could override.

docs/software_requirements/exception_and_error_handling.sdoc
```diff
@@ -64,3 +64,55 @@ The Zephyr RTOS shall provide an interface to assign a specific handler for a fa
...
   VALUE: ZEP-SYRS-5
+
+[REQUIREMENT]
+UID: ZEP-SRS-16-5
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Exception and Error Handling
+TITLE: Triggering a fatal error in the current context
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism for code to trigger a fatal error that terminates the currently running context.
```
anobli, line 75: The function is more complex than what this requirement describe.
The behavior depends on the reason, the context and the handler.
Default handler will halt the system. A custom handler, if it returns, will stop the faulty thread except (and system will keep running) but if the fault happen in the ISR then this treated as a kernel panic and it fails due to an assert.
In my opinion, we need more requirements.

docs/software_requirements/exception_and_error_handling.sdoc
```diff
@@ -64,3 +64,55 @@ The Zephyr RTOS shall provide an interface to assign a specific handler for a fa
...
+  VALUE: ZEP-SYRS-5
+
+[REQUIREMENT]
+UID: ZEP-SRS-16-6
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Exception and Error Handling
+TITLE: Triggering a fatal system error
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism for code to trigger a fatal error that terminates the system.
```
anobli, line 88: I would have used ` halt the system` instead of terminate here.

anobli (commented):
I had to comment requirements outside the scope of this PR. The requirements are in my opinion incorrect and impact how we could read and understand the requirements introduced by this PR. It is worth fixing them.

PR #172, opened 2026-09-06
requirements: add queue cancel waiting thread

nashif:
Sweep of tests/kernel/queue found k_queue_cancel_wait coverage
(test_queue_cancel_wait_error, test_queue_supv_to_user) with no
requirement; queues lacked the equivalent of the FIFO's ZEP-SRS-24-2.
Add ZEP-SRS-20-15 'Queue cancel waiting thread' (parent ZEP-SYRS-22).

### DIFF
--- a/docs/software_requirements/queues.sdoc
+++ b/docs/software_requirements/queues.sdoc
@@ -187,3 +187,16 @@ data items to a queue.
 RELATIONS:
 - TYPE: Parent
   VALUE: ZEP-SYRS-22
+
+[REQUIREMENT]
+UID: ZEP-SRS-20-15
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Queues
+TITLE: Queue cancel waiting thread
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to cause a thread that is blocked waiting to get an item from a queue to return without an item.
+<<<
+RELATIONS:
+- TYPE: Parent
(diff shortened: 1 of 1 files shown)


docs/software_requirements/queues.sdoc
```diff
@@ -187,3 +187,16 @@ data items to a queue.
...
   VALUE: ZEP-SYRS-22
+
+[REQUIREMENT]
+UID: ZEP-SRS-20-15
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Queues
+TITLE: Queue cancel waiting thread
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to cause a thread that is blocked waiting to get an item from a queue to return without an item.
```
anobli, line 198: This is not just a thread, this is the first thread pending in the queue (which means the highest priority thread in queue if we have multiple waiting thread in wait queue)

anobli: That being said, this requirements made me realized that we don't explicitly say that using k_queue_get (ZEP-SRS-20-6 ) could wait. This is implicitly said because we talk about a timeout but I think this makes it ambiguous. Also, same requirement doesn't explain what happens if list is empty and timeout is K_NO_WAIT. And we probably need to explain from ISR, we can't wait (and thus need to set K_NO_WAIT timeout) .
This is outside of scope of this PR,  so its okay to address it later (and maybe open a ticket to track the issue).

PR #173, opened 2026-09-06
requirements: add Futex component

nashif:
The futex fast user-space synchronization primitive (k_futex_wait/
k_futex_wake), exercised by tests/kernel/mem_protect/futex, had no
requirement. Add system requirement ZEP-SYRS-36 (Futex) and a new
Futex software component (ZEP-SRS-37-1..37-3): pending on a futex,
waking pending threads, and user-mode futex operation/access control.

### DIFF
--- a/docs/software_requirements/futex.sdoc
+++ b/docs/software_requirements/futex.sdoc
@@ -1,0 +1,50 @@
+[DOCUMENT]
+TITLE: Futex
+REQ_PREFIX: ZEP-SRS-37-
+
+[GRAMMAR]
+IMPORT_FROM_FILE: software_requirements.sgra
+
+[TEXT]
+STATEMENT: >>>
+SPDX-License-Identifier: Apache-2.0
+<<<
+
+[REQUIREMENT]
+UID: ZEP-SRS-37-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Futex
+TITLE: Pending on a futex
+STATEMENT: >>>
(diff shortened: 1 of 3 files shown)


docs/software_requirements/futex.sdoc
```diff
@@ -0,0 +1,50 @@
...
+<<<
+
+[REQUIREMENT]
+UID: ZEP-SRS-37-1
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Futex
+TITLE: Pending on a futex
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism for a thread to atomically test a futex against an expected value and, if they match, block until the futex is woken or a specified timeout elapses.
```
anobli, line 20: Due to nature of futex and their usage, I would have done two requirements here:
- first one to similar to that one but, saying it return immediately if values differ
- second one starting with a when, saying that when values match, it must block

docs/software_requirements/futex.sdoc
```diff
@@ -0,0 +1,50 @@
...
+  VALUE: ZEP-SYRS-36
+
+[REQUIREMENT]
+UID: ZEP-SRS-37-3
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Futex
+TITLE: Futex operation from user mode
+STATEMENT: >>>
+The Zephyr RTOS shall allow user mode threads to operate on a futex located in memory they can access, and shall reject futex operations on memory the calling thread cannot access.
```
anobli, line 46: we need to requirement, first one, regular case, when thread has access and a second one, starting with if, when we do an invalid access

PR #175, opened 2026-09-06
requirements: clean up and expand Power Management requirements

nashif:
ZEP-SRS-13-2 was a placeholder requirement whose statement was literally
"TBD" and whose title duplicated the component name; remove it. Nothing
referenced it. Also clean the ZEP-SRS-13-1 user story, dropping the
imported "See ZEP104"/"+" artifacts and fixing "an notification" ->
"a notification".

The document then covered only control and notification. Add eleven
requirements (ZEP-SRS-13-4 .. ZEP-SRS-13-14) covering system and device
power management:

- multiple system power states differing in consumption and latency
- entering a low-power state when all CPUs are idle
- power state selection based on anticipated idle time (policy)
- forcing a power state and preventing entry into a power state
- wake-up latency constraints honored by the policy
- device power states (active, suspended, off)
- device power state transitions
- runtime device power management (resume on use, suspend when unused)
- querying a device's current power state
- idling the current CPU until an interrupt occurs

The requirements reflect the implemented system PM (pm_state_force,
pm_policy_*), device PM (pm_device_action_run, pm_device_runtime_*,
pm_device_state_get) and CPU idle (k_cpu_idle, k_cpu_atomic_idle)
behavior following the Route 3s approach and link to the Power
Management system parent (ZEP-SYRS-10). UIDs were generated with
"strictdoc manage auto-uid".

### DIFF
--- a/docs/software_requirements/power_management.sdoc
+++ b/docs/software_requirements/power_management.sdoc
@@ -20,22 +20,7 @@ STATEMENT: >>>
 The Zephyr RTOS shall provide control over changes to system power states.
 <<<
 USER_STORY: >>>
-See ZEP104
-+
-When the power state is changed I want to get an notification in order for specific parts of my application to react accordingly.
-<<<
-RELATIONS:
-- TYPE: Parent
-  VALUE: ZEP-SYRS-10
-
-[REQUIREMENT]
-UID: ZEP-SRS-13-2
-STATUS: Draft
-TYPE: Functional
-COMPONENT: Power Management
-TITLE: Power Management
-STATEMENT: >>>
-TBD
+When the power state is changed I want to get a notification so that specific parts of my application can react accordingly.
 <<<
 RELATIONS:
 - TYPE: Parent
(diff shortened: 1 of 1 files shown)


docs/software_requirements/power_management.sdoc
```diff
@@ -20,35 +20,163 @@ STATEMENT: >>>
 The Zephyr RTOS shall provide control over changes to system power states.
 <<<
 USER_STORY: >>>
-See ZEP104
-+
-When the power state is changed I want to get an notification in order for specific parts of my application to react accordingly.
+When the power state is changed I want to get a notification so that specific parts of my application can react accordingly.
```
anobli, line 23: I don't know which of user story or requirement is wrong, but, they don't seem aligned at all.
Requirement talk about controlling power state and user story is about notification

docs/software_requirements/power_management.sdoc
```diff
@@ -20,35 +20,163 @@ STATEMENT: >>>
...
-UID: ZEP-SRS-13-2
+UID: ZEP-SRS-13-3
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Power Management
-TITLE: Power Management
+TITLE: Notification of changes to system power states
 STATEMENT: >>>
-TBD
+The Zephyr RTOS shall provide notification of changes to system power states.
```
anobli, line 36: This doesn't define anything about pm notifier, except that we want one.
We need requirements to define that we want mechanism to register or unregister pm notifier, also to define what is a pm notifier (entry and exit callback with a power state and substate), when the callback get called and so on.

docs/software_requirements/power_management.sdoc
```diff
@@ -20,35 +20,163 @@ STATEMENT: >>>
...
-UID: ZEP-SRS-13-3
+UID: ZEP-SRS-13-4
 STATUS: Draft
 TYPE: Functional
 COMPONENT: Power Management
-TITLE: Notification of changes to system power states
+TITLE: Multiple system power states
 STATEMENT: >>>
-The Zephyr RTOS shall provide notification of changes to system power states.
+The Zephyr RTOS shall support multiple system power states that differ in power consumption and wake-up latency.
```
anobli, line 49: This is ambiguous. All the states are defined so we could probably define them in requirement here, or as a sub requirements (one for each state, to maybe define precisely the the state).

docs/software_requirements/power_management.sdoc
```diff
@@ -20,35 +20,163 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-10
+
+[REQUIREMENT]
+UID: ZEP-SRS-13-6
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Power Management
+TITLE: Power state selection by anticipated idle time
+STATEMENT: >>>
+The Zephyr RTOS shall select the system power state based on the time remaining until the next scheduled kernel event.
```
anobli, line 75: We should say that is for the default policy.

docs/software_requirements/power_management.sdoc
```diff
@@ -20,35 +20,163 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-10
+
+[REQUIREMENT]
+UID: ZEP-SRS-13-7
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Power Management
+TITLE: Forcing a power state
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to override the power management policy and force the system into a specified power state.
```
anobli, line 88: I fell like we are defining two different things here. And I dislike force word in this context.
We have default policy and we have here, a custom we want to define. In both cases, we expect the policy to return the most appropriated state, and not to force a specific state.

docs/software_requirements/power_management.sdoc
```diff
@@ -20,35 +20,163 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-10
+
+[REQUIREMENT]
+UID: ZEP-SRS-13-8
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Power Management
+TITLE: Preventing entry into a power state
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to prevent the system from entering a specified power state.
```
anobli, line 101: if this requirement is about pm_policy_state_lock_get, then, we also need one for pm_policy_state_lock_put.

docs/software_requirements/power_management.sdoc
```diff
@@ -20,35 +20,163 @@ STATEMENT: >>>
...
+TYPE: Functional
+COMPONENT: Power Management
+TITLE: Entering a low-power state when idle
+STATEMENT: >>>
+When all CPUs become idle, the Zephyr RTOS shall place the system into a low-power state selected by the power management policy.
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-10
+
```
anobli, line 67: Following requirements seems to be about policy
I think having a section for power management, and one for the policy, here, would be nice.

docs/software_requirements/power_management.sdoc
```diff
@@ -20,35 +20,163 @@ STATEMENT: >>>
...
+[REQUIREMENT]
+UID: ZEP-SRS-13-9
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Power Management
+TITLE: Wake-up latency constraint
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to impose a maximum wake-up latency constraint that the power management policy shall respect when selecting a power state.
```
anobli, line 114: I find documentation more obvious: `The system will not enter any power state that would make the system to exceed the given latency value.`
also, does it affect only the default policy or does also affect application policy ?

anobli: we also need a requirement for the API that updates the the latency, one for the API that removes it.
We also have API to be notified when we are updating a latency so many requirement are missing.

docs/software_requirements/power_management.sdoc
```diff
@@ -20,35 +20,163 @@ STATEMENT: >>>
...
+<<<
+RELATIONS:
+- TYPE: Parent
+  VALUE: ZEP-SYRS-10
+
```
anobli, line 119: I think we could also add a new section since following requirement seem to be about device power management.

docs/software_requirements/power_management.sdoc
```diff
@@ -20,35 +20,163 @@ STATEMENT: >>>
...
+[REQUIREMENT]
+UID: ZEP-SRS-13-10
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Power Management
+TITLE: Device power states
+STATEMENT: >>>
+The Zephyr RTOS shall support device power states, including active, suspended, and off.
```
anobli, line 127: I am wondering if we should also have suspending state.

docs/software_requirements/power_management.sdoc
```diff
@@ -20,35 +20,163 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-10
+
+[REQUIREMENT]
+UID: ZEP-SRS-13-11
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Power Management
+TITLE: Device power state transitions
+STATEMENT: >>>
+The Zephyr RTOS shall provide a mechanism to request a device to transition between its power states.
```
anobli, line 140: I think here should detail what happen in case of error since device power management may fail for many errors.

docs/software_requirements/power_management.sdoc
```diff
@@ -20,35 +20,163 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-10
+
+[REQUIREMENT]
+UID: ZEP-SRS-13-12
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Power Management
+TITLE: Runtime device power management
+STATEMENT: >>>
+The Zephyr RTOS shall provide runtime device power management that resumes a device when it is requested for use and suspends it when it is no longer in use.
```
anobli, line 153: We have multiple API (get, put, put_async), and API to enable this feature for a device. We need more requirements.

anobli: Also, I think that runtime pm would deserve it own section.

docs/software_requirements/power_management.sdoc
```diff
@@ -20,35 +20,163 @@ STATEMENT: >>>
...
+  VALUE: ZEP-SYRS-10
+
+[REQUIREMENT]
+UID: ZEP-SRS-13-14
+STATUS: Draft
+TYPE: Functional
+COMPONENT: Power Management
+TITLE: Idling the current CPU
+STATEMENT: >>>
+The Zephyr RTOS shall provide an interface for the calling thread to place the current CPU into a low-power idle state until an interrupt occurs, including a variant that atomically re-enables interrupts as the CPU enters the idle state.
```
anobli, line 179: We describe two different interfaces here, so we need to requirements.

anobli (changes requested):
This PR miss a lot of requirements and should be completed.
Maybe, if we want to avoid too big PR, we may do three PRs, one for system PM, device PM and runtime PM.

